PR #1, opened 2017-01-19
Rename aggregator, add gitignore, add a few CLI options

maurodoglio:


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -128,12 +144,26 @@ object Aggregator {
...
       ssc,
       kafkaParams,
       Set("telemetry"))
 
-    process("/tmp/parquet")(stream)  // TODO: parametrize the job
+
+    val prefix = s"error_aggregates/v1"
+    val outputBucket = opts.outputBucket()
+
+    val path = if(opts.saveLocally()) "/tmp/parquet" else s"s3://${outputBucket}/${prefix}/"
```
vitillo: What about using an option to pass a prefix for the output path? In that way we don't have to distinguish in the code if the path is local or not.

Issue #19, opened 2017-02-10
Support schema evolution


vitillo:
Ideally we would like to have a generic mechanism in which we declare a metric part of the dashboard, which could be an annotation in the metric definition files. In order to support this use-case, we have to verify if it's possible to change the schema of the data being aggregated at run-time. If that fails, we should verify that to the very least the checkpointing logic works as expected when redeploying the application with an updated schema definition.

maurodoglio:
I verified this works without apparent issues. The only downside is that consumers of the parquet partition need to allow schema merging, but that's a fair requirement. [Here](https://gist.github.com/maurodoglio/5eb993cb3ed42157c0bf9cb19f65800a) is a gist of the notebook I used to test this.

PR #31, opened 2017-02-24
Add SLOW_SCRIPT_PAGE_COUNT to ErrorAggregator stats

maurodoglio:
This is one of the metrics suggested by releng.
I will use it to test schema updates as per #19

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
@@ -49,6 +49,7 @@ object ErrorAggregator {
     .add[Int]("PERMISSIONS_SQL_CORRUPTED")
     .add[Int]("DEFECTIVE_PERMISSIONS_SQL_REMOVED")
     .add[Int]("SLOW_SCRIPT_NOTICE_COUNT")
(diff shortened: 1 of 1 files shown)


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -49,6 +49,7 @@ object ErrorAggregator {
     .add[Int]("PERMISSIONS_SQL_CORRUPTED")
     .add[Int]("DEFECTIVE_PERMISSIONS_SQL_REMOVED")
     .add[Int]("SLOW_SCRIPT_NOTICE_COUNT")
+    .add[Int]("SLOW_SCRIPT_PAGE_COUNT")
```
vitillo: The metric is just declared but never updated.

maurodoglio: 🤔

maurodoglio: I'm not sure I follow you. Once a metric is there, it will added to the schema and it will be retrieved by `getCountHistogramValue`, right?

vitillo: Right.

PR #38, opened 2017-05-02
Add options to run in batch mode

fbertsch:
Worked on a cluster. We don't usually have tests for these kind of code paths though.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
@@ -2,15 +2,17 @@ package com.mozilla.telemetry.streaming
 
 import java.sql.{Date, Timestamp}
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -41,6 +59,8 @@ object ErrorAggregator {
     val failOnDataLoss:ScallopOption[Boolean] = opt[Boolean](
       "failOnDataLoss",
       descr = "Whether to fail the query when it’s possible that data is lost.")
+    requireOne(kafkaBroker, from)
```
maurodoglio: I didn't know this was possible! 😹

fbertsch: I did have to look up the full API docs to find it :)

PR #41, opened 2017-05-11
Add case classes representing crash and main pings.

### LINKED ISSUE #39: Add non-main crash rates to MissionControl
This is about adding the various main/content/shutdown/plugin/gpu crash rates to mission control. Originally created on [bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1357793)

maurodoglio:
This PR introduces a few case classes representing crash and main pings.
This classes are then used to extract ping data out of HekaMessages
Fixes #39

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/pings/package.scala
+++ b/src/main/scala/com/mozilla/telemetry/pings/package.scala
@@ -1,3 +1,6 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
 package com.mozilla.telemetry
 
 import java.sql.Timestamp
--- a/src/main/scala/com/mozilla/telemetry/pings/package.scala
+++ b/src/main/scala/com/mozilla/telemetry/pings/package.scala
@@ -98,7 +98,16 @@ package object pings {
                    `payload.keyedHistograms`: JValue,
                    `payload.histograms`: JValue,
                    `payload.info`: JValue
-                 )
+                 ){
+
+    /**
+      * Returns a java Timestamp obj with microseconds resolution.
+      * The source Timestamp field has nanoseconds resolution
+      */
+    def normalizedTimestamp(): Timestamp = {
+      new Timestamp(this.Timestamp / 1000000)
+    }
+  }
 
   case class CrashPayload(
                            crashDate: String,
@@ -118,10 +127,6 @@ package object pings {
 
     def isMainCrash(): Boolean = {
       payload.processType.getOrElse("main") == "main"
-    }
-
-    def timestamp(): Timestamp = {
-      new Timestamp(this.meta.Timestamp / 1000000)
     }
   }
 
@@ -159,10 +164,6 @@ package object pings {
           case _ => 0
         }
       } catch { case _: Throwable => 0 }
-    }
-
-    def timestamp(): Timestamp = {
-      new Timestamp(this.meta.Timestamp / 1000000)
     }
   }
 
--- a/src/main/scala/com/mozilla/telemetry/pings/package.scala
+++ b/src/main/scala/com/mozilla/telemetry/pings/package.scala
@@ -113,7 +113,7 @@ package object pings {
                         meta: Meta
                       ){
 
-    def isMain(): Boolean = {
+    def isMainCrash(): Boolean = {
       payload.processType.getOrElse("main") == "main"
     }
 
--- a/src/main/scala/com/mozilla/telemetry/pings/package.scala
+++ b/src/main/scala/com/mozilla/telemetry/pings/package.scala
@@ -138,32 +138,34 @@ package object pings {
                        // Environment omitted because it's mostly available under meta
                        meta: Meta
                      ){
-    def getCountHistogramValue(histogram_name: String): Int = {
+    def getCountHistogramValue(histogram_name: String): Option[Int] = {
       try {
         this.meta.`payload.histograms` \ histogram_name \ "values" \ "0" match {
-          case JInt(count) => count.toInt
-          case _ => 0
+          case JInt(count) => Some(count.toInt)
+          case _ => None
         }
-      } catch { case _: Throwable => 0 }
+      } catch { case _: Throwable => None }
     }
 
-    def getCountKeyedHistogramValue(histogram_name: String, key: String): Int = {
+    def getCountKeyedHistogramValue(histogram_name: String, key: String): Option[Int] = {
       try {
         this.meta.`payload.keyedHistograms` \ histogram_name \ key \ "values" \ "0" match {
-          case JInt(count) => count.toInt
-          case _ => 0
+          case JInt(count) => Some(count.toInt)
+          case _ => None
         }
-      } catch { case _: Throwable => 0 }
+      } catch { case _: Throwable => None }
     }
 
-
-    def usageHours(): Float = {
+    def usageHours: Option[Float] = {
+      val seconds_per_hour = 3600
+      val max_hours = 25
+      val min_hours = 0
       try {
         this.meta.`payload.info` \ "subsessionLength" match {
-          case JInt(length) => Math.min(25, Math.max(0, length.toFloat / 3600))
-          case _ => 0
+          case JInt(length) => Some(Math.min(max_hours, Math.max(min_hours, length.toFloat / seconds_per_hour)))
+          case _ => None
         }
-      } catch { case _: Throwable => 0 }
+      } catch { case _: Throwable => None }
     }
   }
 
--- a/src/main/scala/com/mozilla/telemetry/pings/package.scala
+++ b/src/main/scala/com/mozilla/telemetry/pings/package.scala
@@ -41,9 +41,11 @@ package object pings {
 
   case class System(os: SystemOs, gfx: Option[SystemGfx])
 
-  case class ActiveExperiment(id: String, branch: String)
+  case class OldStyleExperiment(id: String, branch: String)
 
-  case class Addons(activeExperiment: Option[ActiveExperiment])
+  case class NewStyleExperiment(branch: String)
+
+  case class Addons(activeExperiment: Option[OldStyleExperiment])
 
   case class Settings(
       blocklistEnabled: Option[Boolean],
@@ -85,6 +87,7 @@ package object pings {
       `environment.system`: Option[System],
       `environment.profile`: Option[Profile],
       `environment.addons`: Option[Addons],
+      `environment.experiments`: Option[Map[String, NewStyleExperiment]],
       // Main ping fields preparsed by hindsight
       `payload.simpleMeasurements`: JValue,
       `payload.keyedHistograms`: JValue,
@@ -97,6 +100,29 @@ package object pings {
       */
     def normalizedTimestamp(): Timestamp = {
       new Timestamp(this.Timestamp / 1000000)
+    }
+
+    /**
+      * Returns a Tuple2(experimentId, experimentBranch).
+      *
+      * This information could be in 2 different locations depending on whether the
+      * user is enrolled in a new style experiment or in a old style one.
+      *
+      * To further complicate things, in case of a new style experiment, the user
+      * could be enrolled in more than one experiment. In that case we arbitrary decide
+      * to pick the first experiment.
+      */
+    def experiment: Option[(String, String)] = {
+      val oldStyleExperiment = for {
+        addons <- this.`environment.addons`
+        experiment <- addons.activeExperiment
+      } yield (experiment.id, experiment.branch)
+
+      val newStyleExperiment = for {
+        experiments <- this.`environment.experiments`
+        (experimentId, experiment) <- experiments.headOption
+      } yield (experimentId, experiment.branch)
+      oldStyleExperiment orElse newStyleExperiment
     }
   }
 
--- a/src/main/scala/com/mozilla/telemetry/pings/package.scala
+++ b/src/main/scala/com/mozilla/telemetry/pings/package.scala
@@ -51,8 +51,8 @@ package object pings {
                        isDefaultBrowser: Option[Boolean],
                        e10sEnabled: Option[Boolean],
                        e10sCohort: Option[String],
-                       locale: String,
-                       telemetryEnabled: Boolean
+                       locale: Option[String],
+                       telemetryEnabled: Option[Boolean]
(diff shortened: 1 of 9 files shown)


fbertsch:
Per bug 1334457: What about moving these case classes to moztelemetry, where we can import them into either t-b-v, here, or (eventually) other projects? Could then start to move other case classes there as well.

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -0,0 +1,225 @@
+package com.mozilla.telemetry
+
+import java.sql.Timestamp
+
+import com.mozilla.telemetry.heka.Message
+import org.json4s._
+import org.json4s.jackson.JsonMethods.parse
+
+package object pings {
```
vitillo: Let's move this to the moztelemetry library as suggested by Frank.

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -141,69 +137,69 @@ object ErrorAggregator {
...
+    dimensions("os_name") = meta.`environment.system`.map(_.os.name)
+    dimensions("os_version") = meta.`environment.system`.map(_.os.version)
+    dimensions("architecture") = meta.`environment.build`.flatMap(_.architecture)
+    dimensions("country") = Some(meta.geoCountry)
+    dimensions.build
+  }
+
+  private def parseCrashPing(ping: CrashPing): Tuple1[Row] = {
+    // Non-main crashes are already retrieved from main pings
+    if(!ping.isMain()) return Tuple1(null)
```
vitillo: How can a crash ping be a main ping?

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -141,69 +137,69 @@ object ErrorAggregator {
...
+    dimensions("build_id") = meta.`environment.build`.flatMap(_.buildId)
+    dimensions("application") = Some(meta.appName)
+    dimensions("os_name") = meta.`environment.system`.map(_.os.name)
+    dimensions("os_version") = meta.`environment.system`.map(_.os.version)
+    dimensions("architecture") = meta.`environment.build`.flatMap(_.architecture)
+    dimensions("country") = Some(meta.geoCountry)
+    dimensions.build
+  }
+
+  private def parseCrashPing(ping: CrashPing): Tuple1[Row] = {
```
vitillo: Can we move this functionality within the CrashPing class itself?

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -141,69 +137,69 @@ object ErrorAggregator {
...
+    if(!ping.isMain()) return Tuple1(null)
 
+    val dimensions = buildDimensions(ping.meta)
+    val stats = new RowBuilder(statsSchema)
+    stats("count") = Some(1)
+    stats("main_crashes") = Some(1)
+    Tuple1(RowBuilder.merge(dimensions, stats.build))
+  }
+
+  private def parseMainPing(ping: MainPing): Tuple1[Row] = {
```
vitillo: Can we move this functionality within the MainPing class itself?

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
maurodoglio: I should rename that method to isMainCrash, it's a bit confusing with MainPing around

src/main/scala/com/mozilla/telemetry/pings/package.scala
maurodoglio: While I like the principle, I think this would be a pain to handle in practice for 2 reasons:
1- Changes to a central ping declaration (think of a new non-optional field) will require changes in every place where that ping is used, across different repositories. 
2- Different projects may need different parts of a ping. For example in this patch I kept the definition of the pings intentionally loose. This made it easier to create test fixtures and it likely keeps the memory footprint at a minimum (not too sure about this though).
That said, this is the second time that I spend time defining the structure of a crash ping as a case class, I can see the benefits of moving this to moztelemetry.
@vitillo @fbertsch final thoughts?

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
maurodoglio: This function is very specific to ErrorAggregator, the value it returns is coupled to the output schema.

vitillo: Right, but you could use the "pimp my class" pattern.

maurodoglio: I swear I was going to propose that. I love to pimp my classes 😎

maurodoglio:
RFOL

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -0,0 +1,225 @@
...
+                         )
+
+  case class CrashPing(
+                        application: Application,
+                        clientId: Option[String],
+                        payload: CrashPayload,
+                        // Environment is omitted it's partially available under meta
+                        meta: Meta
+                      ){
+
```
vitillo: Please remove the newline. We should add a linter to travis to have an uniform style.

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -0,0 +1,225 @@
...
+    def isMainCrash(): Boolean = {
```
vitillo: Please add another method for content process crashes.

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -0,0 +1,225 @@
...
+      payload.processType.getOrElse("main") == "main"
+    }
+
+    def timestamp(): Timestamp = {
+      new Timestamp(this.meta.Timestamp / 1000000)
```
vitillo: Please add a comment specifying the timestamp resolution.

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -0,0 +1,225 @@
...
+
+  case class PayloadInfo(subsessionLength: Option[Int])
+
+  case class MainPing(
+                       application: Application,
+                       clientId: Option[String],
+                       // Environment omitted because it's mostly available under meta
+                       meta: Meta
+                     ){
+    def getCountHistogramValue(histogram_name: String): Int = {
```
vitillo: Please add an assert to verify that the desired histogram is in fact a count histogram. The same comment applies elsewhere in this class.

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -0,0 +1,225 @@
...
+      try {
+        this.meta.`payload.histograms` \ histogram_name \ "values" \ "0" match {
+          case JInt(count) => count.toInt
+          case _ => 0
+        }
+      } catch { case _: Throwable => 0 }
```
vitillo: A missing histogram should not generate a value of 0 but something equivalent to "NA" (e.g. null), as the semantics is quite different. The same comment applies elsewhere in this class.

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+
+    def usageHours(): Float = {
+      try {
+        this.meta.`payload.info` \ "subsessionLength" match {
+          case JInt(length) => Math.min(25, Math.max(0, length.toFloat / 3600))
```
vitillo: Why 25 and not 24?

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -0,0 +1,225 @@
...
+                               version: Option[String],
+                               buildId: Option[String],
+                               architecture: Option[String]
+                             )
+
+  case class EnvironmentSystem(os: OS)
+
+  case class OS(
+                 name: Option[String],
+                 version: Option[String])
```
vitillo: The formatting of case classes is inconsistent. E.g. sometimes the closing parentheses is right next to the last field while other times it isn't. Please use a linting tool.

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -0,0 +1,225 @@
...
+
+  def messageToCrashPing(message: Message): CrashPing = {
```
vitillo: Can we make this method a constructor?

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -0,0 +1,225 @@
...
+      "environment.settings",
+      "environment.system",
+      "environment.profile",
+      "environment.addons"
+    )
+    val ping = messageToPing(message, jsonFieldNames)
+    ping.extract[CrashPing]
+  }
+
+  def messageToMainPing(message: Message): MainPing = {
```
vitillo: Can we make this method a constructor?

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -141,6 +131,74 @@ object ErrorAggregator {
...
+    dimensions("submission_date") = Some(new Date(meta.Timestamp / 1000000))
+    dimensions("channel") = Some(meta.normalizedChannel)
+    dimensions("version") = meta.`environment.build`.flatMap(_.version)
+    dimensions("build_id") = meta.`environment.build`.flatMap(_.buildId)
+    dimensions("application") = Some(meta.appName)
+    dimensions("os_name") = meta.`environment.system`.map(_.os.name)
+    dimensions("os_version") = meta.`environment.system`.map(_.os.version)
+    dimensions("architecture") = meta.`environment.build`.flatMap(_.architecture)
+    dimensions("country") = Some(meta.geoCountry)
+    dimensions("experiment_id") = for {
```
vitillo: This works for old style telemetry experiments but Quantum experiments are going to use shield, which supports multiple experiments. @sunahsuh can comment on the right fields to use.

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -141,6 +131,74 @@ object ErrorAggregator {
...
+    dimensions("gfx_compositor") = for {
+      system <- meta.`environment.system`
+      gfx <- system.gfx
+      features <- gfx.features
+      compositor <- features.compositor
+    } yield compositor
+    dimensions.build
+  }
+
+  class ParsableCrashPing(ping: CrashPing) {
```
vitillo: I don't think this is a good name for this class.

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
sunahsuh: Talked to @maurodoglio offline but for posterity, the docs for the new experiment block is here: https://gecko.readthedocs.io/en/latest/toolkit/components/telemetry/telemetry/data/environment.html#experiments

src/main/scala/com/mozilla/telemetry/pings/package.scala
maurodoglio: I'm not sure it's needed because we are only taking main crashes coming from crash pings. Everything else is filtered out to avoid double counting.

src/main/scala/com/mozilla/telemetry/pings/package.scala
maurodoglio: I didn't touch this formula here. This comes from [crash_aggregates](https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/CrashAggregateView.scala#L326), which inherited it from [crash_aggregator](https://github.com/mozilla/moz-crash-rate-aggregates/blob/master/crash_rate_aggregates/crash_aggregator.py#L163). It seems to be an arbitrary cap to normalize the maximum usage hours that a ping can represent. I'm not sure what kind of analysis led to 25 vs 24.

src/main/scala/com/mozilla/telemetry/pings/package.scala
maurodoglio: I reformatted these classes following the scala style guidelines. I also activated scalastyle, but unfortunately it doesn't provide any linting for this kind of problems.

src/main/scala/com/mozilla/telemetry/pings/package.scala
maurodoglio: I moved this method to pings.Meta and added a comment

src/main/scala/com/mozilla/telemetry/pings/package.scala
maurodoglio: Activated scalastyle and fixed the problem

src/main/scala/com/mozilla/telemetry/pings/package.scala
maurodoglio: I changed these methods to return an Option

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
maurodoglio: Changed to ErrorAggregatorCrashPing. I couldn't think of anything shorter and meaningful.

src/main/scala/com/mozilla/telemetry/pings/package.scala
maurodoglio: Done

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
maurodoglio: Done

maurodoglio:
I think I addressed all the issues. @vitillo  this is RFAL!

maurodoglio:
Apparently this version of the job is not producing any data, neither in streaming nor in batch mode. I'm debugging it.

maurodoglio:
I'll `rebase --autosquash` before merging

PR #52, opened 2017-06-09
Add mean firstPaint

maurodoglio:
This is the first measure that requires a non-sum aggregation. I added
a map to define which columns require which aggregation. If a column is
not listed there, it falls back to a sum().


maurodoglio:
I'm having a second thought on this patch... I should store the sum of firstPaint instead of the avg. This will be calculated at query time using sum(firstPaint) / count

maurodoglio:
I left the fixup as a separate commit, I'll autosquash it before merging

maurodoglio:
This is RFAL

PR #53, opened 2017-06-12
Add quantum_ready dimension

maurodoglio:


src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -46,7 +46,36 @@ package object pings {
 
   case class NewStyleExperiment(branch: String)
 
-  case class Addons(activeExperiment: Option[OldStyleExperiment])
+  case class ActiveAddon(isSystem: Option[Boolean], isWebExtension: Option[Boolean]) {
+    def isQuantumReady: Boolean = {
+      this.isSystem orElse this.isWebExtension getOrElse false
```
fbertsch: I treat this differently. If neither `isSystem` or `isWebExtension` are available, then `isQuantumReady` is `null` since it is unknown.

maurodoglio: argh this is wrong. I'll fix it

src/test/scala/com/mozilla/telemetry/streaming/TestPings.scala
```diff
@@ -42,4 +42,85 @@ class TestPings extends FlatSpec with Matchers{
...
+
+  "An Addons instance with invalid Addons and new theme" should "not be quantum ready" in {
+    Addons(Some(invalidAddonMap), None, Some(recentTheme)).isQuantumReady should be (Some(false))
+  }
+
+  "An Addons instance with valid Addons and new theme" should "be quantum ready" in {
+    Addons(Some(validAddonMap), None, Some(recentTheme)).isQuantumReady should be (Some(true))
+  }
+
+  "A Meta instance with e10sEnable and quantumReady addons" should "not be quantumReady" in {
```
fbertsch: The text of this test is not what the test is actually testing, and the actual test is repeated below it. I think it can be removed entirely?

src/main/scala/com/mozilla/telemetry/pings/package.scala
maurodoglio: Yeah I changed it

src/test/scala/com/mozilla/telemetry/streaming/TestPings.scala
maurodoglio: Good catch!

maurodoglio:
This is RFAL

PR #56, opened 2017-06-28
Split window column into 2 timestamp columns

maurodoglio:
This makes the dataset easier to interact with for downstream consumers.


src/test/scala/com/mozilla/telemetry/streaming/TestAggregator.scala
```diff
@@ -94,6 +95,7 @@ class TestAggregator extends FlatSpec with Matchers{
     results("first_paint") should be (42 * 1200)
     results("first_subsession_count") should be (42)
 
-    df.where("window is null").count() should be (0)
+    results("window_start").toString should be ("2016-04-07 14:35:00.0")
+    results("window_end").toString should be ("2016-04-07 14:40:00.0")
```
wlach: From the travis run, it looks like these are wrong?

PR #57, opened 2017-07-17
Add a test for new style experiments

robhudson:
Add test for pref flip experiments which required a parsing update.

### DIFF
--- a/src/test/scala/com/mozilla/telemetry/streaming/TestAggregator.scala
+++ b/src/test/scala/com/mozilla/telemetry/streaming/TestAggregator.scala
@@ -7,19 +7,24 @@ import java.sql.Timestamp
 
 import org.apache.spark.sql.SparkSession
 import org.json4s.DefaultFormats
-import org.scalatest._
+import org.scalatest.{BeforeAndAfterAll, FlatSpec, Matchers}
 
-class TestAggregator extends FlatSpec with Matchers{
+class TestAggregator extends FlatSpec with Matchers with BeforeAndAfterAll {
 
   implicit val formats = DefaultFormats
   val k = TestUtils.scalarValue
   val app = TestUtils.application
 
+  val spark = SparkSession.builder()
+    .appName("Error Aggregates")
+    .master("local[1]")
+    .getOrCreate()
+
+  override def afterAll() {
+    spark.stop()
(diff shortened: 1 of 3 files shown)


src/test/scala/com/mozilla/telemetry/streaming/TestAggregator.scala
```diff
@@ -100,4 +100,55 @@ class TestAggregator extends FlatSpec with Matchers{
     results("window_start").asInstanceOf[Timestamp].getTime should be <= (TestUtils.testTimestampMillis)
     results("window_end").asInstanceOf[Timestamp].getTime should be >= (TestUtils.testTimestampMillis)
   }
-}
+
+  "The aggregator" should "handle new style experiments" in {
+    val spark = SparkSession.builder()
```
fbertsch: Instead of creating spark here, move it to a `beforeAll()` function. Remove it from the other test as well. Add an `afterAll()` function which calls `spark.stop()`. For an example of this, see [spark-hyperloglog tests](https://github.com/mozilla/spark-hyperloglog/blob/master/src/test/scala/com/mozilla/spark/sql/hyperloglog/test/HyperLogLog.scala). Note you need to mixin the `BeforeAndAfterAll` trait.

src/test/scala/com/mozilla/telemetry/streaming/TestAggregator.scala
```diff
@@ -100,4 +100,55 @@ class TestAggregator extends FlatSpec with Matchers{
...
+      )))
+    val messages = (crashMessage ++ mainMessage).map(_.toByteArray).seq
+    val df = ErrorAggregator.aggregate(spark.sqlContext.createDataset(messages).toDF, raiseOnError = true, online = false)
+    df.count() should be (1)
+    val inspectedFields = List(
+      "experiment_id",
+      "experiment_branch"
+    )
+    val row = df.select(inspectedFields(0), inspectedFields.drop(1):_*).first()
+    val results = 0.to(inspectedFields.length-1).map(x => (inspectedFields(x), row(x))).toMap
```
fbertsch: This can be replaced with `row.toSeq.zip(inspectedFields).toMap`, which is more readable.

PR #58, opened 2017-07-24
Add Experiments Aggregates Table

fbertsch:
There's 2 main changes:
1. Remove experiment id and branch from error_aggregates
2. Add new ExperimentErrorAggregates dataset


maurodoglio:
Yesterday we agreed that this is not gonna happen. We are instead going to change error_aggregates to explode the experiment_id column

PR #59, opened 2017-07-27
Use displayVersion instead of version for version field

wlach:
displayVersion has some extra metadata (b12...) which is useful for
distinguishing different betas.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/pings/package.scala
+++ b/src/main/scala/com/mozilla/telemetry/pings/package.scala
@@ -16,6 +16,7 @@ package object pings {
       buildId: String,
       channel: String,
       name: String,
+      displayVersion: String,
       platformVersion: String,
       vendor: String,
       version: String,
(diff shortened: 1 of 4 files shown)


wlach:
@digitarald this is more a heads up than anything else. I don't think you're consuming beta information, but if you were, expect the version field to change from things like "55.0" to "55.0b1" etc. My understanding is that for other channels things should remain the same.

@whd we need this fix to the error_aggregates generator to get more precise beta results -- I believe the number of unique dimensions should more or less the same, as we would already be subdividing uniquely based on buildid.

wlach:
Argh travis is failing :(

wlach:
Ok, travis is fixed now.

wlach:
Hmm, I had heard of buildhub but didn't know it was so far along. I think it should provide the functionality we need to distinguish beta's by build id without changing the schema in this way. Using buildid by itself would likely result in a confusing ui.

I'll leave this open while I explore that option.

wlach:
Closing this in favor of https://github.com/mozilla/missioncontrol/pull/113

wlach:
Reopening this, I think we want this for the mission control dashboard. Except let's add displayVersion as an additional dimension. It shouldn't increase the overall cardinality of the dataset since there should be a 1:1 match between displayVersion and buildId.

wlach:
Actually let's just open a new pr for this: #83.

PR #60, opened 2017-08-04
Bug 1383759 - Experiment aggregates

fbertsch:
See the commits for details.


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -170,48 +170,53 @@ object ErrorAggregator {
...
-      features <- gfx.features
-      compositor <- features.compositor
-    } yield compositor
-    dimensions("quantum_ready") = meta.isQuantumReady
-    dimensions.build
+
+  private def buildDimensions(meta: Meta): Array[Row] = {
+
+    // add a null experiment_id and experiment_branch for each ping
+    val experiments = (meta.experiments :+ (None, None)).toSet.toArray
```
maurodoglio: I'm not 100% sure about adding a null experiment id/branch. If we do so we will not be able to tell which users were not enrolled in any experiment. This is not a blocker for the review but it would be good to have chat about it.

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -132,37 +132,28 @@ package object pings {
...
         addons <- this.`environment.addons`
         experiment <- addons.activeExperiment
-      } yield (experiment.id, experiment.branch)
+      } yield (Some(experiment.id), Some(experiment.branch))
 
-      val newStyleExperiment = for {
-        experiments <- this.`environment.experiments`
-        (experimentId, experiment) <- experiments.headOption
-      } yield (experimentId, experiment.branch)
-      oldStyleExperiment orElse newStyleExperiment
```
fbertsch: The return type is now all experiments, not just the first one.

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
fbertsch: Well basically you're asking a different question than the one this code is designed to answer. If we wanted to know about users who were not enrolled in any experiments, we could add that as a separate dimension (a `boolean`: whether the user is enrolled in an experiment), or add the number of experiments the client is enrolled in as an `int`.

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -92,12 +92,12 @@ object ErrorAggregator {
...
     .add[String]("architecture")
     .add[String]("country")
-    .add[String]("experiment_id")
-    .add[String]("experiment_branch")
     .add[Boolean]("e10s_enabled")
     .add[String]("e10s_cohort")
     .add[String]("gfx_compositor")
     .add[Boolean]("quantum_ready")
+    .add[String]("experiment_id")
+    .add[String]("experiment_branch")
```
fbertsch: No good reason - let me fix this!

fbertsch:
@maurodoglio RFAL

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
maurodoglio: Good point. I'll make sure I'll bring this up in the next MC meeting to see if it's necessary.

PR #61, opened 2017-08-10
Add new profile_age_days dimension

maurodoglio:
This new dimension allows us to distinguish between new profiles, young
profiles, old profiles and so on. I arbitrarily decided that profiles
older than 60 days will all end up in the 60th bucket to avoid an
explosion of the number of aggregates.

I also increased the schema version as this change is not backward
compatible and it will require a data backfill.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
@@ -22,7 +22,7 @@ import org.joda.time.DateTime
 object ErrorAggregator {
 
   private val allowedDocTypes = List("main", "crash")
-  private val outputPrefix = "error_aggregates/v1"
+  private val outputPrefix = "error_aggregates/v2"
   private val kafkaCacheMaxCapacity = 1000
 
   private class Opts(args: Array[String]) extends ScallopConf(args) {
@@ -109,6 +109,7 @@ object ErrorAggregator {
     .add[String]("e10s_cohort")
     .add[String]("gfx_compositor")
     .add[Boolean]("quantum_ready")
+    .add[Int]("profile_age_days")
     .build
 
   private val metricsSchema = new SchemaBuilder()
@@ -236,6 +237,9 @@ object ErrorAggregator {
       dimensions("quantum_ready") = meta.isQuantumReady
       dimensions("experiment_id") = experiment_id
       dimensions("experiment_branch") = experiment_branch
+      dimensions("profile_age_days") = meta.`environment.profile`.flatMap(
+        _.ageDaysBin(new DateTime(meta.normalizedTimestamp().getTime))
+      )
       dimensions.build
     }
   }
--- a/src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
+++ b/src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
@@ -4,8 +4,11 @@
 package com.mozilla.telemetry.streaming
 
 import java.sql.Timestamp
+
 import com.mozilla.spark.sql.hyperloglog.functions.{hllCreate, hllCardinality}
+import com.mozilla.telemetry.streaming.TestUtils.todayDays
 import org.apache.spark.sql.SparkSession
+import org.joda.time.{DateTime, Duration}
 import org.json4s.DefaultFormats
 import org.scalatest.{BeforeAndAfterAll, FlatSpec, Matchers}
 
@@ -73,7 +76,8 @@ class TestErrorAggregator extends FlatSpec with Matchers with BeforeAndAfterAll 
       "window_start",
       "window_end",
       "HllCardinality(client_count) as client_count",
-      "HllCardinality(long_main_input_latency_client_count) as long_main_input_latency_client_count"
+      "HllCardinality(long_main_input_latency_client_count) as long_main_input_latency_client_count",
+      "profile_age_days"
     )
 
     val query = df.selectExpr(inspectedFields:_*)
@@ -118,6 +122,7 @@ class TestErrorAggregator extends FlatSpec with Matchers with BeforeAndAfterAll 
     results("window_end").head.asInstanceOf[Timestamp].getTime should be >= (TestUtils.testTimestampMillis)
     results("client_count") should be (Set(1))
     results("long_main_input_latency_client_count") should be (Set(1))
+    results("profile_age_days") should be (Set(70))
   }
 
   "The aggregator" should "handle new style experiments" in {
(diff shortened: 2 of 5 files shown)


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -20,7 +20,7 @@ import org.joda.time.DateTime
 object ErrorAggregator {
 
   private val allowedDocTypes = List("main", "crash")
-  private val outputPrefix = "error_aggregates/v1"
+  private val outputPrefix = "error_aggregates/v2"
```
maurodoglio: I think it does require a version bump. From my understanding if we change the dimensions schema the aggregations in flux are not valid anymore, so we cannot just restart the job.

maurodoglio:
Good catch with bucket limit. I meant to run it by either @digitarald or @bsmedberg before landing it. Not sure what the requirements are here.

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
fbertsch: In that case it depends on the deploy procedure. If we: store the current position in Kafka, write the current aggregates, stop the job, and then start the new job with the same current position, I don't see any reason why the aggregates wouldn't just keep working.

maurodoglio: This [bug comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1357277#c17) has some info re: the deploy procedure.
Also, Databricks suggest to [start a separate job is](https://databricks.com/blog/2017/05/18/taking-apache-sparks-structured-structured-streaming-to-production.html) in case of a schema change.
I [empirically checked](https://github.com/mozilla/telemetry-streaming/issues/19) that adding measures doesn't require any particular handling, but that's not the case for dimensions.

fbertsch: It seems odd to me that starting a new job is required for some types of schema evolution, but not others. It is definitely not ideal that we would have to bump versions and backfill all of the data on dimension changes.

If this is indeed the limitation, one option is we partition by `submission_date`, start running the new schema, and then copy over the data from the previous `submission_dates`. We can either backfill the overlap day for the new schema, or simply copy over the old schema's data for that day. Then we don't need to backfill, just copy the files on schema evolution compatible schema changes.

maurodoglio: Ok let's move this conversation to [bug 1390522](https://bugzilla.mozilla.org/show_bug.cgi?id=1390522).

maurodoglio:
cc'ing @chutten  as he ran some profile age-based analysis in the past, maybe he has an opinion on this.

maurodoglio:
The current proposal is to use age in weeks with a maximum cap of 52. Stuart to confirm that with someone from product.

stuartphilp:
@maurodoglio Chatted with Product, they do need daily granularity for the first couple of weeks as that is typically where the see the most churn with users. I propose we do daily for the first 14 days, and then weekly for weeks 2 to 52, and the leave it as "> 52 weeks" or "> 1 year" beyond that.

maurodoglio:
From a technical perspective fewer buckets means faster aggregation and faster query execution. It's not easy to give you a precise good/bad number, we will have to give it a try. I'll go for the 89 buckets solution, with daily granularity up to 42 days and then weekly up to 1y.

maurodoglio:
Hey @sunahsuh could you take a quick look at this patch? @fbertsch is on PTO today and it would be great to get another pair of 👀 look at it

sunahsuh:
I can take a look after 1pm today (interview)

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -236,6 +237,7 @@ object ErrorAggregator {
       dimensions("quantum_ready") = meta.isQuantumReady
       dimensions("experiment_id") = experiment_id
       dimensions("experiment_branch") = experiment_branch
+      dimensions("profile_age_days") = meta.`environment.profile`.flatMap(_.ageDaysBin(new DateTime()))
```
sunahsuh: It makes me a bit nervous to tie profile age to the current server clock, in case we ever replay/backfill data -- is there a specific reason we want to use this instead of submission date?

src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
```diff
@@ -170,7 +175,7 @@ class TestErrorAggregator extends FlatSpec with Matchers with BeforeAndAfterAll
     results("experiment_id").toSet should be (Set("new-experiment-1", "new-experiment-2", null))
     results("experiment_branch").toSet should be (Set("control", "chaos", null))
   }
-
+
```
sunahsuh: Nit: whitespace

src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
```diff
@@ -317,8 +322,55 @@ class TestErrorAggregator extends FlatSpec with Matchers with BeforeAndAfterAll
       "HllCardinality(client_count) as client_count",
       "HllCardinality(session_count) as session_count"
     ).collect()(0)
-    results.getAs[Any]("long_main_input_latency_session_count") should be (5)
-    results.getAs[Any]("client_count") should be (1)
-    results.getAs[Any]("session_count") should be (10)
+    results.getAs[Any]("long_main_input_latency_session_count") should be(5)
+    results.getAs[Any]("client_count") should be(1)
+    results.getAs[Any]("session_count") should be(10)
```
sunahsuh: Curious why these are getting caught in the diff..

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
maurodoglio: Good catch!

src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
maurodoglio: right you are :microscope:

src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
maurodoglio: sneaky whitespaces

maurodoglio:
@sunahsuh I addressed your comments :) thanks!

PR #62, opened 2017-08-14
Add client_count-based statistics

maurodoglio:
This change adds the ability to cheaply compute user set cardinality
using hyperloglog.
HLL is used to both count the number of distinct users for each
aggegate and to measure the impact that a certain anomaly has on users,
e.g. how many users are affected by main crashes.


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -149,26 +169,32 @@ object ErrorAggregator {
...
-      fieldName <- statsSchema.fieldNames
-      normFieldName = fieldName.toLowerCase
-    } yield sum(normFieldName).alias(normFieldName)
-
     if (online) {
       parsedPings = parsedPings.withWatermark("timestamp", "1 minute")
     }
 
+    val dimensionsCols = List(
+      col("window.start").as("window_start"),
```
maurodoglio: I moved that change to a preliminary commit

maurodoglio:
@fbertsch I improved the code coverage, do you wanna take another look or should I merge it once Travis is done?

PR #63, opened 2017-08-16
Add session count-based statistics

maurodoglio:
Same as 8687e8e6b861f50f178bfa841bea64b09b00cd36, but this time for
session counts. I added client_id to the recipe for the session hll
to mitigate the risk of collisions.

### DIFF
--- a/src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
+++ b/src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
@@ -54,6 +54,7 @@ class TestErrorAggregator extends FlatSpec with Matchers with BeforeAndAfterAll 
       "gmplugin_crashes",
       "content_shutdown_crashes",
       "count",
+      "subsession_count",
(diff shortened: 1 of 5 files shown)


src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
```diff
@@ -237,4 +250,55 @@ class TestErrorAggregator extends FlatSpec with Matchers with BeforeAndAfterAll
...
+                |      "10000": 0
+                |    }
+                |  }
+                |}""".stripMargin
+          )))
+      )
+
+    val messages = (mainMessagesAffected ++ mainMessagesNotAffected).map(_.toByteArray).seq
+    val df = ErrorAggregator.aggregate(spark.sqlContext.createDataset(messages).toDF, raiseOnError = true, online = false)
+    val session_count = df.selectExpr(
```
fbertsch: Can you confirm in this test that we see: 1 client, and 10 sessions?

PR #64, opened 2017-08-28
Fix runtime error

fbertsch:
I can't prove to you that this fails with a test *yet*, but soon!


maurodoglio:
@fbertsch it may be worth it to spend some time soon to dig into this error. Thanks for the quick patch :)

PR #65, opened 2017-08-29
Discard non-firefox pings on ingestion

maurodoglio:


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -309,6 +309,10 @@ object ErrorAggregator {
     if (!allowedDocTypes.contains(docType)) {
       throw new Exception("Doctype should be one of " + allowedDocTypes.mkString(sep = ","))
     }
+    val appName = fields.getOrElse("appName", "").asInstanceOf[String]
+    if (appName != "Firefox") {
```
fbertsch: Move this to a variable `AllowedAppNames`

PR #67, opened 2017-08-30
Add integration tests

fbertsch:
Alright, this should make it easy to try out new changes with spark-streaming!

This also shows the bug I mentioned.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,21 +15,40 @@ organization := "com.mozilla"
 
 scalaVersion in ThisBuild := "2.11.8"
 
-val sparkVersion = "2.1.1"
+val sparkVersion = "2.2.0"
 
 lazy val root = (project in file(".")).
   settings(
     libraryDependencies += "com.mozilla.telemetry" %% "moztelemetry" % "1.0-SNAPSHOT",
     libraryDependencies += "com.mozilla.telemetry" %% "spark-hyperloglog" % "2.0.0-SNAPSHOT",
-    libraryDependencies += "org.scalatest" %% "scalatest" % "2.2.1" % "test",
+    libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.1" % "test",
     libraryDependencies += "org.apache.spark" %% "spark-core" % sparkVersion % "provided",
     libraryDependencies += "org.apache.spark" %% "spark-streaming" % sparkVersion % "provided",
     libraryDependencies += "org.apache.spark" %% "spark-sql" % sparkVersion % "provided",
     libraryDependencies += "org.apache.spark" %% "spark-sql-kafka-0-10" % sparkVersion,
     libraryDependencies += "org.rogach" %% "scallop" % "1.0.2",
     libraryDependencies += "com.google.protobuf" % "protobuf-java" % "2.5.0",
-    libraryDependencies += "joda-time" % "joda-time" % "2.9.2"
+    libraryDependencies += "joda-time" % "joda-time" % "2.9.2",
+    libraryDependencies += "org.apache.kafka" % "kafka_2.11" % "0.10.0.1"
   )
+
+// Setup docker task
+enablePlugins(DockerComposePlugin, DockerPlugin)
+dockerImageCreationTask := docker.value
+composeFile := sys.props.getOrElse("DOCKER_DIR", default = "docker/") + "docker-compose.yml"
+variablesForSubstitutionTask := {
+    val dockerKafkaHost: String = "./docker_setup.sh" !!;
+    Map("DOCKER_KAFKA_HOST" -> dockerKafkaHost)
+}
+
+// Only run docker tasks on `sbt dockerComposeTest`
+testOptions in Test += Tests.Argument("-l", "DockerComposeTag")
+
+dockerfile in docker := {
+  new Dockerfile {
+    from("java")
+  }
+}
 
 // make run command include the provided dependencies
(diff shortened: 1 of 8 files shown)


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -209,7 +209,7 @@ object ErrorAggregator {
       .withColumn("client_hll", expr("HllCreate(client_id, 12)"))
       .withColumn("session_hll", expr("HllCreate(concat(client_id, session_id), 12)"))
       .groupBy(dimensionsCols: _*)
-      .agg(aggCols(0), aggCols.drop(0): _*)
+      .agg(aggCols.head, aggCols.tail: _*)
```
maurodoglio: 🕵️ good work Frank! 🕵️‍♀️

src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
```diff
@@ -386,6 +424,89 @@ class TestErrorAggregator extends FlatSpec with Matchers with BeforeAndAfterAll
...
+      override def onQueryStarted(event: StreamingQueryListener.QueryStartedEvent): Unit = {
+        sendRecords()
+      }
+
+      override def onQueryProgress(event: StreamingQueryListener.QueryProgressEvent): Unit = {
+        messagesSeen += event.progress.numInputRows
+
+        // Spark Bug? Sending messages after getting QueryStartedEvent are still missed
+        // Sending again if the query hasn't seen any
+        if(messagesSeen == 0){
```
maurodoglio: So this function should run on every trigger, right? That means we keep sending it at the same interval of the trigger, but it takes at least 30 seconds to finish. Unless this is run synchronously, but I don't think so.

fbertsch:
@maurodoglio alright, no sleeps! Also it doesn't work to use `zookeeper:2181` and `kafka:9092`, oh well :/

Next up: Travis build

fbertsch:
@maurodoglio, @wlach would either of you kindly try and run this branch locally? Just checkout the branch, and run `sbt dockerComposeTest`. I'm running out of clues for what's wrong - the kafka log indicates it is receiving the message being sent by the test, but for some reason Spark is not reading them.

fbertsch:
@maurodoglio, @wlach now it works :)

fbertsch:
@maurodoglio All ready for another look 👁 👁

src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
```diff
@@ -4,22 +4,39 @@
...
+import scala.collection.JavaConversions._
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+class TestErrorAggregator extends AsyncFlatSpec with Matchers with BeforeAndAfterAll {
 
-class TestErrorAggregator extends FlatSpec with Matchers with BeforeAndAfterAll {
+  object DockerComposeTag extends Tag("DockerComposeTag")
+
+  val zkHostInfo = "localhost:2181"
```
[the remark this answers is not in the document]
maurodoglio: hah, I just noticed that only kafka and zookeeper are running on docker, not sbt/spark.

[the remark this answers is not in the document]
fbertsch: Ah right, it would work if we were running everything in docker. We can consider that down the line if compatibility becomes an issue.

build.sbt
```diff
@@ -15,22 +15,43 @@ organization := "com.mozilla"
...
+composeFile := sys.props.getOrElse("DOCKER_DIR", default = "docker/") + "docker-compose.yml"
+variablesForSubstitutionTask := {
+    val dockerKafkaHost: String = "./docker_setup.sh" !!;
+    Map("DOCKER_KAFKA_HOST" -> dockerKafkaHost)
+}
+
+// Only run docker tasks on `sbt dockerComposeTest`
+testOptions in Test += Tests.Argument("-l", "DockerComposeTag")
+
+dockerfile in docker := {
```
maurodoglio: I'm not sure I understand what's going on here

src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
```diff
@@ -264,7 +303,91 @@ class TestErrorAggregator extends FlatSpec with Matchers with BeforeAndAfterAll
...
+    def send(rs: Seq[Array[Byte]]): Unit = {
+      rs.foreach{ v =>
+        val record = new ProducerRecord[String, Array[Byte]](ErrorAggregator.kafkaTopic, v)
+        kafkaProducer.send(record)
+        kafkaProducer.flush()
+      }
+    }
+
+    val earlier = (TestUtils.generateMainMessages(k, timestamp=Some(1459994516000000000L)) ++
+      TestUtils.generateCrashMessages(k, timestamp=Some(1459994516000000000L))).map(_.toByteArray)
```
maurodoglio: it would be great to have this timestamp defined somewhere in the test class as a constant

build.sbt
fbertsch: Apparently I didn't totally understand either :)

`docker` is an sbt `task`. We are setting the `dockerfile`; `Docker` is the class where we can add steps to the `dockerfile`. Here all of our steps are defined by `docker-compose`, but we can add others (using scalaized versions of the typical `Docker` commands, like `RUN`).

PR #69, opened 2017-09-14
Set number of files per partition

fbertsch:
Previously was just doing 1.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
@@ -26,6 +26,11 @@ object ErrorAggregator {
   private val outputPrefix = "error_aggregates/v2"
   private val kafkaCacheMaxCapacity = 1000
 
+  // This is the number of files output per submission_date
+  // in a batch run.
+  // 180GB total per day, so ~300MB per parquet file.
+  private val numPartitions = 600
+
   private class Opts(args: Array[String]) extends ScallopConf(args) {
     val kafkaBroker: ScallopOption[String] = opt[String](
       "kafkaBroker",
@@ -375,6 +380,7 @@ object ErrorAggregator {
     val outputPath = opts.outputPath()
 
     aggregate(pingsDataframe, raiseOnError = opts.raiseOnError(), online = false)
+      .repartition(numPartitions)
       .write
       .mode("overwrite")
       .partitionBy("submission_date")


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -375,6 +376,7 @@ object ErrorAggregator {
     val outputPath = opts.outputPath()
 
     aggregate(pingsDataframe, raiseOnError = opts.raiseOnError(), online = false)
+      .repartition(numPartitions)
```
maurodoglio: So here we say we want 600 partitions and below we say we want to partition by day. That means we will likely have empty partitions unless we run the batch job on 600 days, right?

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -25,6 +25,7 @@ object ErrorAggregator {
   private val allowedAppNames = List("Firefox")
   private val outputPrefix = "error_aggregates/v2"
   private val kafkaCacheMaxCapacity = 1000
+  private val numPartitions = 600
```
maurodoglio: Where does 600 come from?

fbertsch: right now: 180GB total. 180 / 600 = ~300MB per file, a good size for parquet.

Obviously this will need to be updated when the job runs :)

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
fbertsch: Nope, this gives 600 partitions per day.

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
maurodoglio: Do you mind to add that as a comment?

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
maurodoglio: Is it 600 partitions in total or 600 per day?

fbertsch: Proof :)

```
Welcome to Scala 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_102).
Type in expressions for evaluation. Or try :help.

scala> import org.apache.spark.sql.SparkSession
import org.apache.spark.sql.SparkSession

scala> val spark = SparkSession.builder().master("local[1]").getOrCreate()
17/09/15 08:51:20 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
spark: org.apache.spark.sql.SparkSession = org.apache.spark.sql.SparkSession@349a6105

scala> import spark.implicits._
import spark.implicits._

scala> List(("a", 1), ("a", 2), ("b", 1), ("b", 2)).toDF("id", "val")
res0: org.apache.spark.sql.DataFrame = [id: string, val: int]

scala> List(("a", 1), ("a", 2), ("b", 1), ("b", 2)).toDF("id", "val").repartition(2).write.partitionBy("id").parquet("/tmp/test-output")
[Stage 1:>                                                          (0 + 1) / 2]SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.

scala> import scala.sys.process._
import scala.sys.process._

scala> "ls -R /tmp/test-output"!
warning: there was one feature warning; re-run with -feature for details
_SUCCESS
id=a
id=b

/tmp/test-output/id=a:
part-00000-59e8f480-6975-4527-89f1-3316cf42a186.snappy.parquet
part-00001-59e8f480-6975-4527-89f1-3316cf42a186.snappy.parquet

/tmp/test-output/id=b:
part-00000-59e8f480-6975-4527-89f1-3316cf42a186.snappy.parquet
part-00001-59e8f480-6975-4527-89f1-3316cf42a186.snappy.parquet
res8: Int = 0
```

maurodoglio: This is great, thanks!

PR #72, opened 2017-09-15
Remove session-based counts and filtered client counts

maurodoglio:
We found out that a handful of the HLL fields introduced to handle session-based counts and filtered client counts killed the dimension of the output parquet files. 
See https://bugzilla.mozilla.org/show_bug.cgi?id=1388351 for more context.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
@@ -89,15 +89,6 @@ object ErrorAggregator {
     "CYCLE_COLLECTOR_MAX_PAUSE" -> (List("main", "content"), List(150, 250, 2500))
   )
 
-  private val clientCounts = Map(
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -190,15 +189,13 @@ object ErrorAggregator {
 
     val stats = statsSchema.fieldNames.map(_.toLowerCase)
     val sumCols = stats.map(s => sum(s).as(s))
-    val countCols = clientCounts.flatMap{
+    val countCols = clientCounts map {
```
fbertsch: I thought we were going to remove all `HLL` cols except `client_count`?

maurodoglio: Right you are!

maurodoglio:
This is now RFAL

PR #76, opened 2017-09-20
Limit build ids to the last 6 months

maurodoglio:
This will reduce the number of aggregates we generate, which should
speed up aggregation and reduce the size of the output parquet file.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/pings/package.scala
+++ b/src/main/scala/com/mozilla/telemetry/pings/package.scala
@@ -34,7 +34,7 @@ package object pings {
       xpcomAbi: String)
 
   case class SystemOs(name: String, version: String) {
-    def normalizedVersion: String = OS(Option(name), Option(version)).normalizedVersion
+    val normalizedVersion: String = OS(Option(name), Option(version)).normalizedVersion
   }
 
   case class SystemGfxFeatures(compositor: Option[String])
@@ -162,7 +162,7 @@ package object pings {
       new Timestamp(this.Timestamp / 1000000)
     }
 
-    def normalizedBuildId(): Option[String] = {
+    val normalizedBuildId: Option[String] = {
       `environment.build`.flatMap(_.buildId) match {
         case Some(buildId: String) => {
           val buildIdDay = buildId.slice(0, 8).toString()
@@ -365,7 +365,7 @@ package object pings {
 
   case class OS(name: Option[String], version: Option[String]){
     val versionRegex = "(\\d+(\\.\\d+)?(\\.\\d+)?)?.*".r
-    def normalizedVersion: String = {
+    val normalizedVersion: String = {
       version match {
         case Some(v) =>
           val versionRegex(normalized, b, c) = v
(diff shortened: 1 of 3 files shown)


src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -161,6 +161,19 @@ package object pings {
       new Timestamp(this.Timestamp / 1000000)
     }
 
+    def normalizedBuildId(): Option[String] = {
+      `environment.build`.flatMap(_.buildId) match {
+        case Some(buildId: String) => {
+          val buildIdDay = buildId.slice(0, 6).toString()
```
fbertsch: This is taking the `YYYYMM` of the build, you need `YYYYMMDD`

src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
```diff
@@ -425,4 +425,28 @@ class TestErrorAggregator extends AsyncFlatSpec with Matchers with BeforeAndAfte
     val df = ErrorAggregator.aggregate(spark.sqlContext.createDataset(messages).toDF, raiseOnError = false, online = false)
     df.schema.fields.map(_.name) should not contain ("client_id")
   }
+
+  "BuildId" should "not be older than 6 months" in {
+    import spark.implicits._
+    val messages = TestUtils.generateMainMessages(
+      1, Some(Map(
+        "environment.build" -> """{"buildId": "20170602"""",
```
fbertsch: These tests are testing the case of a build after a submission date. We should be testing the opposite - a build BEFORE a submission date; specifically: a build < 6 months before a submission_date, and a build > 6 months before a submission date. Any builds after a submission_date should be ignored, since that makes no sense!

src/main/scala/com/mozilla/telemetry/pings/package.scala
maurodoglio: Right you are

src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
maurodoglio: Good catch, I also found another bug while fixing this 😄

maurodoglio:
This is RFAL

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -159,6 +160,24 @@ package object pings {
       new Timestamp(this.Timestamp / 1000000)
     }
 
+    def normalizedBuildId(): Option[String] = {
+      `environment.build`.flatMap(_.buildId) match {
+        case Some(buildId: String) => {
+          val buildIdDay = buildId.slice(0, 8).toString()
+          val buildDateFormat = DateTimeFormat.forPattern("yyyyMMdd")
+          val buildDateTime = buildDateFormat.parseDateTime(buildIdDay)
```
fbertsch: This will fail on improper dates which our schema validation [0] doesn't necessarily catch. For example, `00000000` is validated, but it will cause this to fail with `org.joda.time.IllegalFieldValueException: Cannot parse "00000000": Value 0 for monthOfYear must be in the range [1,12]`.

[0] https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/schemas/telemetry/main/main.4.schema.json#L11

PR #78, opened 2017-09-22
Add experiments_aggregates job

fbertsch:
This is exactly error_aggregates, but with a subset of
dimensions and statistics.

The only other changes made were to enable missing schema elements.


fbertsch:
@maurodoglio I *think* it will be easier for mission-control related integration to keep them in `error_aggregates`. Let's consider this a temporary measure until we get that up and running.

fbertsch:
Also, the build is failing :( working on it

PR #80, opened 2017-10-10
Stream events to amplitude

fbertsch:


src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
```diff
@@ -0,0 +1,224 @@
...
+    verify()
+  }
+
+  def parsePing(message: Message, sample: Double): Array[String] = {
+    implicit val formats = DefaultFormats
+
+    val fields = message.fieldsAsMap
+    val docType = fields.getOrElse("docType", "").asInstanceOf[String]
+    if (!allowedDocTypes.contains(docType)) {
+      throw new Exception("Doctype should be one of " + allowedDocTypes.mkString(sep = ","))
```
sunahsuh: I'm curious -- the default kafka topic is "telemetry", and this the firstdocType/appName check I see in the streaming path -- are they filtered before this?

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
```diff
@@ -0,0 +1,224 @@
...
+
+  def main(args: Array[String]): Unit = {
+    val opts = new Opts(args)
+
+    val spark = SparkSession.builder()
+      .appName(queryName)
+      .getOrCreate()
+
+    opts.kafkaBroker.get match {
+      case Some(_) => storeStreamingEvents(spark, opts)
```
sunahsuh: Slight naming nit: we're not actually writing out in these "store*" methods, correct? It seems like "send" or something like it would be more appropriate

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
```diff
@@ -0,0 +1,224 @@
...
+      "startingOffsets",
+      descr = "Starting offsets (streaming mode only)",
+      required = false,
+      default = Some("latest"))
+    val url: ScallopOption[String] = opt[String](
+      "url",
+      descr = "Endpoint to send data to",
+      required = true)
+    val apiKey: ScallopOption[String] = opt[String](
+      "apiKey",
```
sunahsuh: passing the api key as a command-line param (and, therefore, having it in the bash history, etc) makes me a little nervous, tbh, but that's whd's domain

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -376,7 +389,43 @@ package object pings {
...
+        val newEvents = currentEvents match {
+          case JArray(x) => JArray(x.map{ e =>
+            e match {
+              case JArray(event) =>
+                JObject(
+                  JField("timestamp", event(0))         ::
+                  JField("category", event(1))          ::
+                  JField("method", event(2))            ::
+                  JField("object", event(3))            ::
+                  JField("value", event.lift(4).getOrElse(JNull)) ::
```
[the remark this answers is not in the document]
fbertsch: Yes, I couldn't agree more! See https://github.com/mozilla/telemetry-streaming/pull/41#issuecomment-300860424

src/main/scala/com/mozilla/telemetry/pings/FocusEvent.scala
```diff
@@ -0,0 +1,64 @@
...
+    compact(render(eventsList))
+  }
+
+  def getHashedClientId: String = {
+    val messageDigest = MessageDigest.getInstance("SHA-256")
+    new String(messageDigest.digest(clientId.getBytes()))
+  }
+
+  def includeClient(sample: Double, seed: Int = 42): Boolean = {
+    val randish = (MurmurHash3.stringHash(clientId, seed) & 0x7FFFFFFF) / Int.MaxValue.toDouble
```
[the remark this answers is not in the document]
fbertsch: Hmm, I'm not sure. I'll need to chat with IT to get an idea of what kind of sampling we should do.

src/main/scala/com/mozilla/telemetry/pings/FocusEvent.scala
```diff
@@ -0,0 +1,64 @@
...
+      ("os_version" -> osversion) ~
+      ("country" -> meta.geoCountry) ~
+      ("city" -> meta.geoCity)
+    }
+
+    compact(render(eventsList))
+  }
+
+  def getHashedClientId: String = {
+    val messageDigest = MessageDigest.getInstance("SHA-256")
```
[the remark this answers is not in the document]
fbertsch: Yup, in the case that their data is exposed, it can't be tracked back to a clientId on our end.

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
fbertsch: I'll make sure that we're using best practice for this.

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
fbertsch: Nope, we don't know what the docType is until we unpack the protobuf, so it can't be done until this point. There has been talk of per-docType topics though.

sunahsuh: Then wouldn't the streaming version fail with an exception immediately?

fbertsch: See line 140; we catch the exceptions.

fbertsch:
@sunahsuh RFAL

sunahsuh:
Hmm, travis didn't report status back but looks like the build is still failing, boo

fbertsch:
@sunahsuh tests now passing, also moved loading the API key via env var.

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
sunahsuh: Hmm, I'm a bit torn on this -- it's not very idiomatic in JVM languages to catch expected exceptions, and it's supposed to be pretty slow to boot. I see the issue re: handling batch vs streaming exceptions differently, but I'd advocate for adding a check on whether we're running in a streaming context or not before throwing an exception as opposed to catching upstream.

In the process of doing that, you'd probably remove the code to catch Throwables, but for future reference: https://www.sumologic.com/blog/code/why-you-should-never-catch-throwable-in-scala/

fbertsch: Good point. We should change this in ErrorAggregator as well. The downside is there are a whole host of errors that can arise during parsing, and it's nigh impossible to decipher what they are and enumerate those during testing. Let me consider what to do here.

fbertsch: @sunahsuh I've avoided catching expected exceptions, but kept the catch-all try. I'm thinking we circle back to this later and fix it in both this code and ErrorAggregator.

PR #81, opened 2017-10-11
Various changes to make this package compile with sbt v1.0

wlach:


project/plugins.sbt
```diff
@@ -1,12 +1,14 @@
-addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.0")
+resolvers += "bintray-spark-packages" at "https://dl.bintray.com/spark-packages/maven/"
```
fbertsch: Is this line still required?

PR #83, opened 2017-10-12
Add displayVersion to error aggregates table

wlach:
This column will give more precise versioning information (e.g. 47.0b1 as
opposed to just 47.0). It should not increase the cardinality of the dataset,
as there should be a 1:1 mapping between it and buildId.


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -209,7 +210,7 @@ object ErrorAggregator {
       .coalesce(1)
   }
 
-  private def buildDimensions(meta: Meta): Array[Row] = {
+  private def buildDimensions(application: Application, meta: Meta): Array[Row] = {
```
fbertsch: nit: make application the second argument

wlach:
Made requested modification

PR #87, opened 2017-11-01
Bug 1411397 - Handle pings with no displayVersion

wlach:


src/test/scala/com/mozilla/telemetry/TestUtils.scala
```diff
@@ -86,14 +86,15 @@ object TestUtils {
     }
   }
 
-  def generateMainMessages(size: Int, fieldsOverride: Option[Map[String, Any]]=None, timestamp: Option[Long]=None): Seq[Message] = {
+  def generateMainMessages(size: Int, fieldsOverride: Option[Map[String, Any]]=None, timestamp: Option[Long]=None,
+                           removeDisplayVersion: Boolean=false): Seq[Message] = {
```
fbertsch: Can we change this to a `removeFields: List[String]`, which are fields to remove? The match below would just look like `case JField(x, _) if removeFields.contains(x) => true`.

wlach: Sure, I did this way because there is only one field we are currently removing.

wlach: I wound up naming the variable 'fieldsToRemove' instead

wlach:
Updated per your request, could you have another look?

src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
```diff
@@ -449,4 +449,18 @@ class TestErrorAggregator extends AsyncFlatSpec with Matchers with BeforeAndAfte
     val df2 = ErrorAggregator.aggregate(spark.sqlContext.createDataset(messages).toDF, raiseOnError = false, online = false)
     df2.where("build_id IS NULL").collect().length should be (0)
   }
+
+  "display version" can "be undefined" in {
+    import spark.implicits._
+    val messages = TestUtils.generateMainMessages(
+      1, None, None, List[String]("displayVersion")
```
fbertsch: nit: you don't have to specify the `String` type here, since it can infer it from the function definition. So you can use either `List("displayVersion")`, or my personal favorite, `"displayVersion" :: Nil`

wlach: 👍

PR #90, opened 2017-11-06
Bug 1414942 - Move object variables to function arguments

fbertsch:
Spark distributes the objects at the beginning of the run. That
means that any updates to the object only occur on the master node.
This means previously, the schema didn't match between what
the master node expected and the slaves wrote.

This moves all those object variables to function arguments,
which are guaranteed to be evaluated at runtime, not distribution
time.


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -33,7 +33,6 @@ object ErrorAggregator {
   private val allowedAppNames = List("Firefox")
   private val kafkaCacheMaxCapacity = 1000
 
-
```
wlach: Did you intend to make this change?

PR #93, opened 2017-12-07
Batch run by day

fbertsch:


fbertsch:
@wlach this is a permanent change. This is not for `error_aggregates` but for `experiment_error_aggregates`, so the current real-time streaming will remain unaffected. This change simply ensures that backfills won't overwrite existing data.

wlach:
Looking at context and discussing on irc, this looks ok to me, but as discussed I may not understand this codebase (or scala) well enough to be the best reviewer.

PR #96, opened 2018-02-05
Backport displayVersion schema addition

wlach:
This is the "v1" branch with the code to add a displayVersion column backported. Hopefully we can deploy this successfully while waiting for a solution that allows us to deploy the rest of the functionality in the master branch, since it unblocks a pretty important relman use case -- using mission control to monitor beta rollouts.

### DIFF
--- a/src/test/scala/com/mozilla/telemetry/streaming/TestAggregator.scala
+++ b/src/test/scala/com/mozilla/telemetry/streaming/TestAggregator.scala
@@ -14,12 +14,13 @@ class TestAggregator extends FlatSpec with Matchers{
   implicit val formats = DefaultFormats
   val k = TestUtils.scalarValue
   val app = TestUtils.application
+  val spark = SparkSession.builder()
+    .appName("Error Aggregates")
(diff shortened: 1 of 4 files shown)


src/test/scala/com/mozilla/telemetry/streaming/TestAggregator.scala
```diff
@@ -15,6 +15,12 @@ class TestAggregator extends FlatSpec with Matchers{
   val k = TestUtils.scalarValue
   val app = TestUtils.application
 
+  val spark = SparkSession.builder()
```
whd: It looks to me like this initialization should either be moved into the new test case (line 113), or the similar initialization logic at line 25 should be removed.

fbertsch: Agreed, we don't need it per-test.

wlach:
I'm going to merge this with the suggested fix to the unit tests. @whd feel free to create a new PR with whatever you changes you want to apply against the v1a branch.

PR #99, opened 2018-03-19
Refactor tests to conform with Scalastyle rules, enable Scalastyle in test scope on CI

akkomar:
This enables Scalastyle checks for tests. Also, some small refactors were needed so all tests conform to the rules.

### DIFF
--- a/src/test/scala/com/mozilla/telemetry/TestUtils.scala
+++ b/src/test/scala/com/mozilla/telemetry/TestUtils.scala
@@ -86,6 +86,7 @@ object TestUtils {
     }
   }
 
+  // scalastyle:off methodLength
   def generateMainMessages(size: Int, fieldsOverride: Option[Map[String, Any]]=None, timestamp: Option[Long]=None,
                            fieldsToRemove: List[String] = List[String]()): Seq[Message] = {
     val defaultMap = Map(
@@ -195,6 +196,7 @@ object TestUtils {
       )
     }
   }
+  // scalastyle:on methodLength
 
   def generateFocusEventMessages(size: Int, fieldsOverride: Option[Map[String, Any]]=None, timestamp: Option[Long]=None): Seq[Message] = {
     val defaultMap = Map(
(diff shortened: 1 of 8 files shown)


src/test/scala/com/mozilla/telemetry/streaming/TestEventsToAmplitude.scala
```diff
@@ -92,6 +92,7 @@ class TestEventsToAmplitude extends FlatSpec with Matchers with BeforeAndAfterAl
   override def afterEach {
     verify(expectedTotalMsgs,
       requestMadeFor(new ValueMatcher[Request] {
+        // scalastyle:off methodName
```
fbertsch: It seems we should instead update the regex to allow this kind of method name, with both a preceding and following back-tick.

src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
```diff
@@ -386,7 +386,9 @@ class TestErrorAggregator extends FlatSpec with Matchers {
 
       override def onQueryTerminated(event: StreamingQueryListener.QueryTerminatedEvent): Unit = {
         if(messagesSeen < expectedTotalMsgs){
+          // scalastyle:off print-ln
```
fbertsch: We can instead remove this `println`, it hasn't really been useful since some initial debugging.

src/test/scala/com/mozilla/telemetry/TestUtils.scala
```diff
@@ -86,6 +86,7 @@ object TestUtils {
     }
   }
 
+  // scalastyle:off methodLength
```
fbertsch: I'm open to shortening this.

src/test/scala/com/mozilla/telemetry/streaming/TestEventsToAmplitude.scala
akkomar: +1

src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
akkomar: +1

src/test/scala/com/mozilla/telemetry/TestUtils.scala
akkomar: I'll leave it like this for now. As far as I understand this part these util methods would lose a little on readability if refactored for shortening.

PR #101, opened 2018-03-19
Bug 1423340 - Make tests easily runnable from IDE

akkomar:
Previously it wasn't simple to easily and repeatedly execute tests from IDE.
From now on:
* tests are responsible for cleaning up their state (previously it was docker setup script's job)
* generic option for discovering internal IP address is added in docker setup script (useful when one's interface names don't match those listed in script)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,3 +24,17 @@ the source code and running the tests via sbt. Some common invocations for sbt:
 * `sbt dockerComposeTest  # run the docker compose tests (slow)`
 * `sbt "dockerComposeTest -tags:DockerComposeTag" # run only tests with DockerComposeTag (while using docker)`
 * `sbt ci  # run all tests`
+
+Some tests need Kafka to run. If one prefers to run them via IDE, it's required to run the test cluster:
+```bash
+sbt dockerComposeUp
+```
+or via plain docker-compose:
+```bash
+export DOCKER_KAFKA_HOST=$(./docker_setup.sh)
+docker-compose -f docker/docker-compose.yml up
+```
+It's also good to shut down the cluster afterwards:
+```bash
+sbt dockerComposeStop
+```
--- a/src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
@@ -3,25 +3,20 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 package com.mozilla.telemetry.streaming
 
-import com.mozilla.telemetry.heka.Message
-import com.mozilla.telemetry.pings._
-import com.mozilla.telemetry.timeseries._
-import com.mozilla.telemetry.streaming.sinks.HttpSink
-import org.apache.spark.sql.types.{BinaryType, StructField, StructType}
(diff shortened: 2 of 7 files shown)


src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
```diff
@@ -202,7 +202,7 @@ object EventsToAmplitude {
 
   def sendStreamingEvents(spark: SparkSession, opts: Opts): Unit = {
     val config = readConfigFile(opts.configFilePath())
-    val apiKey = sys.env(AMPLITUDE_API_KEY_KEY)
+    val apiKey = sys.env.getOrElse(AMPLITUDE_API_KEY_KEY, "test-api-key")
```
fbertsch: This changes production behavior: if the key is missing from the env, it will no longer error out. I'm open to removing the setting of the API key in `runtests.sh` if it can be added only during testing. If you can figure out an alternative for setting it, then we can remove `runtests.sh` altogether.

README.md
```diff
@@ -24,3 +24,17 @@ the source code and running the tests via sbt. Some common invocations for sbt:
 * `sbt dockerComposeTest  # run the docker compose tests (slow)`
 * `sbt "dockerComposeTest -tags:DockerComposeTag" # run only tests with DockerComposeTag (while using docker)`
 * `sbt ci  # run all tests`
+
+Some tests need Kafka to run. If one prefers to run them via IDE, it's required to run test cluster:
```
fbertsch: nit: "required to run the test cluster"

akkomar: +1

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
akkomar: I see. Assuming we don't want to change the way we provide key, how about:
* introducing new method accepting `Opts` and api key to act as entrypoint for processing
* calling this method from `main` with parsed `Opts` and key extracted from environment variable
* calling this method in test (instead of main with list of arguments)
* likewise, refactor `ErrorAggregator`

fbertsch: @akkomar that sounds like a reasonable change to me, and should keep the test api key string consistent within the testing module.

README.md
```diff
@@ -24,3 +24,17 @@ the source code and running the tests via sbt. Some common invocations for sbt:
...
 * `sbt "dockerComposeTest -tags:DockerComposeTag" # run only tests with DockerComposeTag (while using docker)`
 * `sbt ci  # run all tests`
+
+Some tests need Kafka to run. If one prefers to run them via IDE, it's required to run the test cluster:
+```bash
+sbt dockerComposeUp
+```
+or via plain docker-compose:
+```bash
+export DOCKER_KAFKA_HOST=$(ip route get 8.8.8.8 | awk '{print $NF; exit}')
```
fbertsch: I would instead have them run the `docker_setup.sh`, since this command doesn't work on macos.

fbertsch:
@akkomar please rebase these into 1 commit before merging.

README.md
akkomar: Thanks for checking that!

PR #104, opened 2018-03-27
Make HTTP sink more resilient to errors

fbertsch:
This handles:
- Retrying after timeouts
- Retrying after server errors
- Logging on failures

docs here on response codes: https://amplitude.zendesk.com/hc/en-us/articles/204771828#http-status-codes-retrying-failed-requests

Note that this change does not include dealing with too-large payloads (response 413) since at this point we're nowhere near that number for any payload. Bug 1449092 will handle that.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/sinks/HTTPSink.scala
+++ b/src/main/scala/com/mozilla/telemetry/sinks/HTTPSink.scala
@@ -5,8 +5,24 @@ package com.mozilla.telemetry.streaming.sinks
 
 import scalaj.http.{Http, HttpRequest}
 import org.apache.spark.sql.ForeachWriter
+import scala.annotation.tailrec
+import scala.util.{Try, Success, Failure}
 
-class HttpSink[String](url: String, parameters: Map[String, String]) extends ForeachWriter[String] {
+class HttpSink[String](url: String, parameters: Map[String, String], maxAttempts: Int = 5, defaultDelay: Int = 500, connectionTimeout: Int = 2000)
+  extends ForeachWriter[String] {
+
+  // codes to retry a request on - allow retries on timeouts
+  // docs: https://amplitude.zendesk.com/hc/en-us/articles/204771828#http-status-codes-retrying-failed-requests
+
+  val TimeoutPseudoCode = -1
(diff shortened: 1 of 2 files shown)


src/main/scala/com/mozilla/telemetry/sinks/HTTPSink.scala
```diff
@@ -17,8 +33,32 @@ class HttpSink[String](url: String, parameters: Map[String, String]) extends For
...
+
+  @tailrec
+  private def attempt(request: HttpRequest, tries: Int = 0): Unit = {
+    if(tries > 0) { // minor optimization
+      java.lang.Thread.sleep(backoff(tries))
+    }
+
+    val code = Try(request.asString.code) match {
+      case Success(c) => c
+      case Failure(e: java.net.SocketTimeoutException) => TimeoutPseudoCode
```
akkomar: Can we get any other exception here?
If so, we won't send events remaining in current batch.

src/main/scala/com/mozilla/telemetry/sinks/HTTPSink.scala
```diff
@@ -5,8 +5,24 @@ package com.mozilla.telemetry.streaming.sinks
...
+  val RetryCodes = TimeoutPseudoCode :: 429 :: 500 :: 502 :: 503 :: 504 :: Nil
+  val OK = 200
+
+  // timeouts in ms
+  val ReadTimeout = 5000
+
+  @transient lazy val log = org.apache.log4j.LogManager.getLogger("HttpSinkLogger")
 
-class HttpSink[String](url: String, parameters: Map[String, String]) extends ForeachWriter[String] {
   def close(errorOrNull: Throwable): Unit = {}
```
akkomar: How about logging errors here?
We're logging some failed requests @L60, we might lose some error information here.

src/main/scala/com/mozilla/telemetry/sinks/HTTPSink.scala
fbertsch: Added catching and logging any `NonFatal` exceptions. I think we want fatal ones to follow through, since it will cause the job to fail (assuming the error reproduces on retries of the partition), in which case we can investigate.

akkomar: You're right, I didn't think of `Fatal` ones.

PR #107, opened 2018-03-27
Add Android Focus eventsToAmplitude config

fbertsch:


fbertsch:
@sunahsuh yeah it will be a pain. I wrote a parser that takes a pretty easy-to-ready csv file that looks something like:

|  Amplitude Name | Category | Method | Object | Value | Extras | Amplitude Properties |
|:------------- |:-------------:|:-----:|:------------- |:-------------:|:-----:|:-----:|
| Link Context Menu - Item selected | action | share, copy | browser_contextmenu | link | | item=method |
| Start session | action | foreground | app | | | |
| Stop session| action | background | app | | | |

We could consider formalizing that structure.

PR #111, opened 2018-04-10
Use repartition instead of coalesce in streaming job

akkomar:
Coalescing to 1 partition constrains all aggregations to one node, which eventually causes OOMs on production workloads.


akkomar:
Batch experiments job looks good - there was no significant change of duration today.

PR #114, opened 2018-04-13
Add Preferences to Focus Android Events

fbertsch:
This includes:
- Fixing the preferences in the tests to be bools rather than strings
- New case class to represent the different preferences
- Adding the preferences to the Amplitude events


fbertsch:
@sunahsuh `user_properties` is for Amplitude -- thanks for the review!

PR #119, opened 2018-04-23
Setup travis ci deploy

fbertsch:


.travis.yml
```diff
@@ -1,27 +1,67 @@
...
+  - mkdir $JAR_DIR
+  - mv target/scala-2.11/telemetry-streaming-assembly-0.1-SNAPSHOT.jar "$JAR_DIR/$JAR_NAME"
+
+deploy:
+  - provider: s3
+    bucket: telemetry-artifacts-stage
+    local_dir: $JAR_DIR
+    skip_cleanup: true
+    region: us-west-2
+    secret_access_key:
```
jasonthomas: Can these encrypted values be moved into environment variables?

.travis.yml
```diff
@@ -1,27 +1,67 @@
...
+before_deploy:
+  - sbt assembly
+  - export JAR_DIR="telemetry-streaming"
+  - export JAR_NAME="telemetry-streaming.jar"
+  - mkdir $JAR_DIR
+  - mv target/scala-2.11/telemetry-streaming-assembly-0.1-SNAPSHOT.jar "$JAR_DIR/$JAR_NAME"
+
+deploy:
+  - provider: s3
+    bucket: telemetry-artifacts-stage
```
jasonthomas: Can we deploy this to `net-mozaws-data-us-west-2-ops-mavenrepo` similar to what we do for telemetry-batch-view?

.travis.yml
jasonthomas: I was thinking about setting some env vars to the secret values above but I am not sure if that would even work. I think it's fine the way it is, I am just paranoid about having these values (even encrypted) available.

Did you provision a new user + api key for this? If not just let me know and I can help.

fbertsch:
@whd assuming the creds work as expected with s3 deploy, this should give us what we want. I made a few minor changes to the layout, e.g. instead of `latest.jar` and `latest-release.jar`, it's just `telemetry-streaming.jar`.

whd:
This branch looks like it may need a rebase. I'll add the variables presently.

fbertsch:
Rebased.

whd:
For testing purposes, can you set the "on" branch to this branch?

Additionally, I don't see anything about environment variables on https://docs.travis-ci.com/user/deployment/s3/ similar to https://docs.travis-ci.com/user/uploading-artifacts/, so if the build doesn't result in artifacts I would suggest switching to using the artifacts tool.

PR #120, opened 2018-04-24
Make focus settings strings

fbertsch:
Not documented, but they are actually all strings.


fbertsch:
@sunahsuh it's a good suggestion, but in general the custom json4s components are more difficult to follow, especially pre-extraction. This is a very straightforward mapping.

PR #122, opened 2018-04-25
Bug 1453485 - Calculate content_crashes and content_shutdown_crashes using Crash ping's ipc_channel_error

akkomar:
As noted in the ticket and related discussion, this changes meaning of existing fields (`content_crashes` and `content_shutdown_crashes`).


akkomar:
After this is merged, crash ping's schema should be updated: https://github.com/akkomar/mozilla-pipeline-schemas/commit/4e8492332b85607be44b1cc7e0059d15832ced5a

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.1.1
+sbt.version=1.1.4
```
wlach: Did you mean to change this as part of this commit?

akkomar: Yes (it's a small bugfix bump, I'd just push it here)

PR #123, opened 2018-04-26
Use deploy.sh file from telemetry-batch-view

fbertsch:
Already tested.


fbertsch:
@whd tests passed

fbertsch:
To be fair, there was some intermittent failure :)

PR #124, opened 2018-04-26
Bug 1393144 - Use submission_date_s3 for partitioning error_aggregates

akkomar:
Also in separate commit: bump Spark version to 2.3.0 in order to use dynamic partition overwrite.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,8 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-import sys.process._;
+
+import scala.sys.process._;
 
 val localMavenHttps = "https://s3-us-west-2.amazonaws.com/net-mozaws-data-us-west-2-ops-mavenrepo/"
 
@@ -19,7 +20,7 @@ organization := "com.mozilla"
 
 scalaVersion in ThisBuild := "2.11.8"
 
-val sparkVersion = "2.2.0"
+val sparkVersion = "2.3.0"
 
 lazy val root = (project in file(".")).
   settings(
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -29,7 +30,7 @@ lazy val root = (project in file(".")).
     libraryDependencies += "org.apache.spark" %% "spark-core" % sparkVersion % "provided",
     libraryDependencies += "org.apache.spark" %% "spark-streaming" % sparkVersion % "provided",
     libraryDependencies += "org.apache.spark" %% "spark-sql" % sparkVersion % "provided",
-    libraryDependencies += "org.apache.spark" %% "spark-sql-kafka-0-10" % sparkVersion,
+    libraryDependencies += "org.apache.spark" %% "spark-sql-kafka-0-10" % sparkVersion exclude("net.jpountz.lz4", "lz4"),
```
fbertsch: Can you add a comment about the exclusion?

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -379,7 +379,8 @@ object ErrorAggregator {
         .repartition(opts.numParquetFiles())
         .write
         .mode("overwrite")
-        .parquet(s"${outputPath}/${outputPrefix}/submission_date_s3=${currentDate.toString(dateFormat)}")
+        .partitionBy("submission_date_s3")
+        .parquet(s"${outputPath}/${outputPrefix}")
```
fbertsch: This will overwrite everything in `{outputPath}/${outputPrefix}`, deleting old data. Unless we use the new write mode, the `overwrite` mode needs to be used on the specific partition.

build.sbt
akkomar: +1

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
akkomar: Yes, I hardcoded it to use new write mode in https://github.com/mozilla/telemetry-streaming/pull/124/files#diff-e9e9e1ae5916120d8ace633ccc23891cR404
There's also a check that'll fail the job on previous Spark version (since the dynamic mode wasn't there yet): https://github.com/mozilla/telemetry-streaming/pull/124/files#diff-e9e9e1ae5916120d8ace633ccc23891cR407

fbertsch: Ah great, I totally missed that.

PR #126, opened 2018-04-28
Add ability to skip tests

fbertsch:
This will support the quick-development use-case. Tests cannot
be skipped on tagged releases or master.

For example, if an engineer is working on a branch, and wants
to test out her changes using Airflow, they would need to build
the artifact and push it to s3 using a non-master branch. With
tests, that could take upwards of 1 hour -- with [skip-tests],
it should take under 5 minutes.

Note that travis creates a merge commit for PR builds,
so those will always run.


.travis.yml
```diff
@@ -29,7 +29,12 @@ env:
   - DOCKER_DIR="../docker/"
 
 script:
-  - sbt ci
+  - export TEST_SKIP_REGEX="\[skip-tests\]"
+  - if [[ (! $TRAVIS_COMMIT_MESSAGE =~ $TEST_SKIP_REGEX) || -n $TRAVIS_TAG || $TRAVIS_BRANCH == "master" ]];
```
haroldwoo: Shouldn't we only skip tests if the commit message is set?  This conditional implies that we know all the cases in which the tests should run

.travis.yml
```diff
@@ -38,8 +43,19 @@ before_deploy:
...
   - sbt assembly
+  - git config --local user.name "Auto Deployer"
+  - git config --local user.email "telemetry-alerts@mozilla.com"
+  - git tag "$(date +'%Y%m%d%H%M%S')-$(git log --format=%h -1)"
 
 deploy:
+  - provider: releases
+    skip_cleanup: true
+    api_key:
+      secure: Kac5/sXg9TM8cBHrwNch78UiVTl4aA2KQ9jnv0vipUzi341XfmO3EbdEnSCWN+ew39Pc+tAR+nOB+e/AHbOyFjJ5gggA3Z/3UWPYse/8iIPcRs8GC55KfprDrVkD8HO3pPzOXOyDtn64Y6z3cQXGj110QKfPjG6v+LHzNDSl1oxZB7f8bnoNEom5WxZCMAvZO3odbrnKPP333gPpjtmXfe24WlPqBMg3FN+RJMy1lz+bjlDB0hd1qbVdUO3DYivodcn/q1Sdp98EipwCWnOq3JqrmfDhf9rRL1UnSC+WsbcxJL73ee36EkNORSTYVFzhfoSRmNlhqNgGI8DdDhO654fKecXPiWDHyZw+cF/DQHsvmSk5zr5VobeqyByM9cqFj+5gs3RN3CI2UaUbceRwuerwpuFCAVt2WHSw2heYeOlmoLNX2+HFYPCXFJOkzQ5pZl5fEsc280riU/9wlvNfLhKTIJNyP3n3OlU/YlU0FYfR2jF637KAhLX9z8LxY1aEcdcAI7sWCz5jyMGRwvCS7prRneg1ouPn+bUSab7+GMi8F3Rdf1qWZx8IivpXiOw3MubsLLBeXep+PbE70n6AFB68ThgKEAhJsJpqtZKbqpz3w+w6iOe9Ig85P+WdXTK7zC4itWzZfEihk8TTGzerlSHgpMR9lBgvkzJkFY3MQqI=
```
haroldwoo: Is this safe to leave here?

fbertsch: It's a travis encrypted key: https://docs.travis-ci.com/user/encryption-keys/

PR #127, opened 2018-04-29
Use proper submission date format in normalized build_id extraction

akkomar:
This fixes `normalizedBuildId`, which used incorrect date format.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/pings/package.scala
+++ b/src/main/scala/com/mozilla/telemetry/pings/package.scala
@@ -6,8 +6,9 @@ package com.mozilla.telemetry
 import java.sql.Timestamp
 
 import com.mozilla.telemetry.heka.Message
+import com.mozilla.telemetry.pings.Meta._
+import org.joda.time.Months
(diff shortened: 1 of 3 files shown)


src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -143,18 +144,15 @@ package object pings {
...
-
-          p.getMonths() match {
-            case p if 0 to 6 contains(p) => Some(buildId)
+        case Some(buildId: String) =>
+          val buildIdDay = buildId.slice(0, 8)
+          val buildDateTime = BuildDateFormat.parseDateTime(buildIdDay)
+          val submissionDateTime = SubmissionDateFormat.parseDateTime(submissionDate)
+
+          Months.monthsBetween(buildDateTime, submissionDateTime).getMonths match {
+            case _@m if (0 <= m) && (m <= 6) => Some(buildId)
```
fbertsch: Why do it this way instead of `case m if (0 <= m) && (m <= 6) ...`?

PR #128, opened 2018-04-30
Include config files in built JAR

fbertsch:


fbertsch:
Ah right, that old schema can be removed.

We could, but these aren't resources used by the code directly. They are only packaged with the JAR for convenience. I'd opt for keeping them out of the `src/main/resources` directory

akkomar:
OK, so what's using these schemas?

fbertsch:
@akkomar see the related commit here: https://github.com/mozilla/telemetry-airflow/pull/278

akkomar:
@fbertsch I don't know, it seems overcomplicated. You're unzipping the jar and passing it to `EventsToAmplitude` in https://github.com/mozilla/telemetry-airflow/pull/278 anyway. I'd pack it in resources and default to this config file in `EventsToAmplitude`, so:
1) Job definition in https://github.com/mozilla/telemetry-airflow/pull/278 would be simpler
2) It's still possible to pass some arbitrary schemas

But, I might be missing something about this job or bikeshedding a little, so r+ and feel free to decide which way is better.

PR #129, opened 2018-05-02
Bug 1438258 - Track Fennec/Android crashes and usage hours

akkomar:

PR #130, opened 2018-05-02
Ping structure refactor

akkomar:
As a follow up to https://github.com/mozilla/telemetry-streaming/pull/129, this is an attempt to make it more clear which fields are available in each ping.

Due to the way Hindsight parses messages, `meta` may be missing some fields for pings that are not fully in line with common format (e.g. `environment` and `application` objects).
In this commit:
* `Ping` trait is introduced
* concrete ping classes are moved to separate files
* information about presence of certain sections is encoded in ping types using `HasEnvironment` and `HasApplication` traits, which contain extractor methods with implementations specific to these sections
* generic `ErrorAggregator#buildDimensions` method handles all ping types

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/pings/CorePing.scala
+++ b/src/main/scala/com/mozilla/telemetry/pings/CorePing.scala
@@ -1,0 +1,49 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package com.mozilla.telemetry.pings
+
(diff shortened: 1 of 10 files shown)


fbertsch:
@akkomar we should clean this up and chat about it once the immediate work around error_aggregates v2 wraps up. Let me know!

akkomar:
@fbertsch sure. I'll close this and open another PR when I'm done, as it has no longer anything to do with Fennec.

I rebased all the changes to https://github.com/akkomar/telemetry-streaming/tree/pings_refactor. I'd like to also handle validation in a cleaner way with this, but generally it shouldn't change much.
I'd say it's already ready to be discussed, the more so I think it touches problems you mention in https://bugzilla.mozilla.org/show_bug.cgi?id=1461808

PR #131, opened 2018-05-08
Bug 1454047 - Remove unused metrics from error_aggregates

akkomar:
Following fields are removed since they're not used and not needed in v2:
* `subsession_count`
* `first_subsession_count`
* `input_event_response_coalesced_ms_main_above_150`
* `input_event_response_coalesced_ms_main_above_250`
* `input_event_response_coalesced_ms_main_above_2500`
* `input_event_response_coalesced_ms_content_above_150`
* `input_event_response_coalesced_ms_content_above_250`
* `input_event_response_coalesced_ms_content_above_2500`
* `ghost_windows_main_above_1`
* `ghost_windows_content_above_1`
* `gc_max_pause_ms_2_main_above_150`
* `gc_max_pause_ms_2_main_above_250`
* `gc_max_pause_ms_2_main_above_2500`
* `gc_max_pause_ms_2_content_above_150`
* `gc_max_pause_ms_2_content_above_250`
* `gc_max_pause_ms_2_content_above_2500`
* `cycle_collector_max_pause_main_above_150`
* `cycle_collector_max_pause_main_above_250`
* `cycle_collector_max_pause_main_above_2500`
* `cycle_collector_max_pause_content_above_150`
* `cycle_collector_max_pause_content_above_250`
* `cycle_collector_max_pause_content_above_2500`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,7 +25,6 @@ val sparkVersion = "2.3.0"
 lazy val root = (project in file(".")).
   settings(
     libraryDependencies += "com.mozilla.telemetry" %% "moztelemetry" % "1.0-SNAPSHOT",
-    libraryDependencies += "com.mozilla.telemetry" %% "spark-hyperloglog" % "2.0.0-SNAPSHOT",
     libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.1" % Test,
     libraryDependencies += "org.apache.spark" %% "spark-core" % sparkVersion % "provided",
     libraryDependencies += "org.apache.spark" %% "spark-streaming" % sparkVersion % "provided",
--- a/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
@@ -5,8 +5,6 @@ package com.mozilla.telemetry.streaming
 
 import java.sql.Timestamp
 
-import com.mozilla.spark.sql.hyperloglog.aggregates._
-import com.mozilla.spark.sql.hyperloglog.functions._
 import com.mozilla.telemetry.heka.{Dataset, Message}
 import com.mozilla.telemetry.pings._
 import com.mozilla.telemetry.timeseries._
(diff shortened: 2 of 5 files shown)


wlach:
client counts are useful for debugging/understanding the data, please leave those in.

fbertsch:
@wlach have you switched to accessing the data through Presto then?

wlach:
@fbertsch no, still using athena. Maybe we're thinking of something different? here's the exact query missioncontrol is running:

https://github.com/mozilla/missioncontrol/blob/596afd470c3558fea2e0becf43421a08742c7f43/missioncontrol/etl/measure.py#L90

fbertsch:
@wlach ah, that is actually checking that the subsession_count is great than `min_client_count`. To access the `client_count` column, you need to use Presto (or Spark) and cast it as an `HLL` type; e.g. `SELECT submission_date_s3, cardinality(merge(cast(hll AS HLL))) FROM error_aggregates GROUP BY submission_date_s3`

If you're not using it we should remove it, looking at one parquet file shows that it is taking up 86.9% of the parquet file size (even after a compression rate of 17.83!)

wlach:
@fbertsch oh ok! yes, I think we can remove the `client_count` column then. Arguably we might not need even need the subsession count with some changes I have coming down the pipe to use buildhub (that restriction on the query was mainly to filter out bogus builds).

akkomar:
@wlach @fbertsch - you're mentioning subsession count, but I hope you mean submission count here - subsession count is being removed in this PR (and is not used in https://github.com/mozilla/missioncontrol/blob/596afd470c3558fea2e0becf43421a08742c7f43/missioncontrol/etl/measure.py#L90)

akkomar:
@fbertsch I don't think they're the same - `count` is also increased at each crash ping.

fbertsch:
Hmm; @wlach which one do you want to filter on? Total ping count or subsession count?

wlach:
@fbertsch I think total ping count makes more sense for missioncontrol's purposes.

akkomar:
Cool, let's bid farewell to subsession_count and others.

PR #132, opened 2018-05-17
Bug 1453675 - Refactor ping structure

akkomar:
This splits pings' package into multiple files, one per ping type.
Information about presence of certain subobjects is encoded in ping types using `HasEnvironment` and `HasApplication` traits, which contain extractor methods with implementations specific to these objects.

With this change it was also possible to generify `ErrorAggregator#buildDimensions` to handle all ping types.

I'm leaving validation for separate PR as it is job-specific.


src/main/scala/com/mozilla/telemetry/pings/Ping.scala
```diff
@@ -0,0 +1,193 @@
...
+
+  def getOsName: Option[String]
+
+  def getOsVersion: Option[String]
+
+  def getArchitecture: Option[String]
+
+  protected def getRawBuildId: Option[String]
+
+  // TODO: check if we can use appBuildId for all pings
```
fbertsch: We should be able to (for telemetry pings), since it's part of the edge server spec [0].

[0] https://docs.telemetry.mozilla.org/concepts/pipeline/http_edge_spec.html#postput-request

src/main/scala/com/mozilla/telemetry/pings/package.scala
```diff
@@ -3,416 +3,11 @@
...
-import com.mozilla.telemetry.pings.Meta._
-import org.joda.time.Months
-import org.joda.time.format.DateTimeFormat
-import org.json4s.JsonDSL._
 import org.json4s._
 import org.json4s.jackson.JsonMethods._
 
-import scala.util.{Success, Try}
-
 package object pings {
```
fbertsch: We could consider renaming this package object now.

src/main/scala/com/mozilla/telemetry/pings/FocusEvent.scala
```diff
@@ -3,65 +3,86 @@
...
 import com.mozilla.telemetry.heka.Message
 import com.mozilla.telemetry.streaming.EventsToAmplitude.Config
-
-import org.json4s._
 import org.json4s.JsonDSL._
+import org.json4s._
 import org.json4s.jackson.JsonMethods._
 
-import com.fasterxml.jackson.databind.{JsonNode, ObjectMapper}
+case class Event(timestamp: Int,
```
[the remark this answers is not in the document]
sunahsuh: Mine's already implemented as a SendsEventsToAmplitude trait, so I'll just rebase on this when it's ready to go and make the changes I need ;)

src/main/scala/com/mozilla/telemetry/pings/CorePing.scala
```diff
@@ -0,0 +1,49 @@
...
+  override def getOsName: Option[String] = Option(os)
+
+  override def getOsVersion: Option[String] = Option(osversion)
+
+  override def getArchitecture: Option[String] = Option(arch)
+
+  override protected def getRawBuildId: Option[String] = Option(meta.appBuildId)
+
+  def usageHours: Option[Float] = {
+    val seconds_per_hour = 3600
```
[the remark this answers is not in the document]
akkomar: +1

src/main/scala/com/mozilla/telemetry/pings/Ping.scala
akkomar: I'll postpone this for now, there are some crash pings that have these fields different.

src/main/scala/com/mozilla/telemetry/pings/CrashPing.scala
```diff
@@ -0,0 +1,66 @@
...
+      "environment.profile",
+      "environment.addons",
+      "environment.experiments"
+    )
+    val ping = messageToPing(message, jsonFieldNames)
+    ping.extract[CrashPing]
+  }
+}
+
+case class CrashMetadata(StartupCrash: Option[String],
```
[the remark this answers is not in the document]
akkomar: Thanks, it's not consistent indeed.
I haven't seen any guidelines on that. I'll use Spark codebase as a model and reformat FocusEvent to top-to-bottom approach.

src/main/scala/com/mozilla/telemetry/pings/package.scala
akkomar: How about getting rid of the package object?
Since these conversion methods are used only in concrete ping classes, I moved them to `Ping` companion object.

PR #133, opened 2018-05-17
Bug 1461773 - Use core pings's `displayVersion` field for `display_version` dimension

akkomar:
If `displayVersion` is not available, we'll fall back to `meta.appVersion`.

### DIFF
--- a/src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
+++ b/src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
@@ -243,6 +243,29 @@ class TestErrorAggregator extends FlatSpec with Matchers with BeforeAndAfterAll 
     results("build_id") should be(Set(TestUtils.defaultFennecApplication.buildId))
   }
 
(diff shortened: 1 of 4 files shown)


src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
```diff
@@ -243,6 +243,29 @@ class TestErrorAggregator extends FlatSpec with Matchers with BeforeAndAfterAll
...
+      ErrorAggregator.defaultDimensionsSchema, ErrorAggregator.defaultMetricsSchema,
+      ErrorAggregator.defaultCountHistogramErrorsSchema)
+
+    val expectedNumberOfAggregatedRows = 2
+    aggregates.count() shouldBe expectedNumberOfAggregatedRows
+
+    val inspectedFields = List(
+      "display_version"
+    )
+    val rows = aggregates.select(inspectedFields(0), inspectedFields.drop(1): _*).collect()
```
fbertsch: nit: this looks like copy and pasted code from above that could possibly be moved to a function.

PR #134, opened 2018-05-29
Bug 1464191 - Silence logs that are too verbose

relud:
I left this warning in place because it looks something that shouldn't be silenced in prod
<pre>
18/05/29 13:43:37 WARN SharedState: URL.setURLStreamHandlerFactory failed to set FsUrlStreamHandlerFactory
</pre>

### DIFF
--- a/src/main/resources/log4j.properties
+++ b/src/main/resources/log4j.properties
@@ -12,14 +12,20 @@ log4j.logger.org.apache.spark.repl.Main=WARN
 # Settings to quiet third party logs that are too verbose
 log4j.logger.org.spark_project.jetty=WARN
 log4j.logger.org.spark_project.jetty.util.component.AbstractLifeCycle=ERROR
+log4j.logger.org.apache.hadoop.util.NativeCodeLoader=ERROR
 log4j.logger.org.apache.spark.repl.SparkIMain$exprTyper=INFO
 log4j.logger.org.apache.spark.repl.SparkILoop$SparkILoopInterpreter=INFO
+log4j.logger.org.apache.spark.SparkContext=ERROR
+log4j.logger.org.apache.spark.sql.SparkSession$Builder=ERROR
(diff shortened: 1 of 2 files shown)


src/test/scala/com/mozilla/telemetry/streaming/TestErrorAggregator.scala
```diff
@@ -404,18 +405,26 @@ class TestErrorAggregator extends FlatSpec with Matchers with BeforeAndAfterAll
...
-
-    // 1 for each experiment (there are 2), and one for a null experiment
-    df.where("display_version IS NULL").collect().length should be (3)
-    // should be no cases where displayVersion is null for this test case
-    df.where("display_version <> NULL").collect().length should be (0)
+    // silence intentional http failed logs
+    val httpSinkLogger = Logger.getLogger("HttpSinkLogger")
+    val prevLogLevel = httpSinkLogger.getLevel
+    httpSinkLogger.setLevel(Level.ERROR)
+    try {
```
relud: I changed this log level in the code instead of `log4j.properties` because it seems like these kind of warnings would be useful in other contexts where they aren't intentional: 
<pre>
18/05/29 13:46:52 WARN HttpSinkLogger: Failed request: http://localhost:9876/httpapi?k=v, last status code: 666
</pre>

fbertsch:
@sunahsuh yes, all of them are except for the jetty server `IllegalStateException`. It looks like that results from trying to send another response code after one has already been sent. It isn't reproduced locally, I'll make a bug and we can tackle that later.

sunahsuh:
Sounds like in that case we should up the log level in those tests as well for HttpSinkLogger

relud:
okay, I've corrected the log levels again. I left the warnings from wiremock under `HttpChannel` and `Response` in place, because those both appear to be related to the `IllegalStateException`

PR #136, opened 2018-05-30
Fix log output in Http Sink

fbertsch:
This will fix the `k=v` parts of those logs you noticed.


src/main/scala/com/mozilla/telemetry/sinks/HTTPSink.scala
```diff
@@ -64,7 +64,7 @@ class HttpSink[String](url: String, data: Map[String, String], maxAttempts: Int
       case (ErrorPseudoCode, _) =>
       case (c, false) if RetryCodes.contains(c) => attempt(request, tries + 1)
       case (c, _) => {
-        val url = request.url + "?" + request.params.map{ case(k, v) => s"k=v" }.mkString("&")
+        val url = request.url + "?" + request.params.map{ case(k, v) => s"$k=$v" }.mkString("&")
```
sunahsuh: This was a real bug, it looks like -- can you write a regression test for it as well?

fbertsch: Not really - it's just for logging.

PR #137, opened 2018-06-01
Extract common parts of streaming jobs to single base class

akkomar:
Encouraged by: https://github.com/mozilla/telemetry-streaming/pull/135#discussion_r191896546, this extracts common parts of streaming jobs to common base class.
In order to make `ErrorAggregator` and `ExperimentsErrorAggregator` better fit this approach, they are now subclassing their own common base class, which makes their method signatures a little simpler.

As a bonus, joda-time dependency is removed (java.time is used instead).


src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregatorBase.scala
```diff
@@ -0,0 +1,296 @@
...
+      descr = "Output path",
+      required = false,
+      default = Some("/tmp/parquet"))
+    val raiseOnError: ScallopOption[Boolean] = opt[Boolean](
+      "raiseOnError",
+      descr = "Whether the program should exit on a data processing error or not.")
+    val failOnDataLoss: ScallopOption[Boolean] = opt[Boolean](
+      "failOnDataLoss",
+      descr = "Whether to fail the query when it’s possible that data is lost.")
+    val numParquetFiles: ScallopOption[Int] = opt[Int](
```
fbertsch: This is the same as `fileLimit` in the `BaseOpts` class. Why don't we remove this argument in favor of `fileLimit` and move the `conflicts` to `BaseOpts` as well?

src/main/scala/com/mozilla/telemetry/streaming/StreamingJobBase.scala
```diff
@@ -0,0 +1,93 @@
...
+    val kafkaBroker: ScallopOption[String] = opt[String](
+      "kafkaBroker",
+      descr = "Kafka broker (streaming mode only)",
+      required = false)
+    val startingOffsets: ScallopOption[String] = opt[String](
+      "startingOffsets",
+      descr = "Starting offsets (streaming mode only)",
+      required = false,
+      default = Some("latest"))
+
```
fbertsch: nit: inconsistent whitespace

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregatorBase.scala
```diff
@@ -0,0 +1,296 @@
...
+
+import com.mozilla.telemetry.heka.{Dataset, Message}
+import com.mozilla.telemetry.pings.{CorePing, CrashPing, MainPing, Ping}
+import com.mozilla.telemetry.timeseries.{RowBuilder, SchemaBuilder}
+import org.apache.spark.sql.{DataFrame, Row, SparkSession}
+import org.apache.spark.sql.catalyst.encoders.RowEncoder
+import org.apache.spark.sql.functions.{col, sum, window}
+import org.apache.spark.sql.types.{BinaryType, StructField, StructType}
+import org.json4s.DefaultFormats
+import org.rogach.scallop.{ScallopConf, ScallopOption}
```
fbertsch: `ScallopConf` appears to be unused.

src/main/scala/com/mozilla/telemetry/streaming/StreamingJobBase.scala
akkomar: My intention was to group similar options together. After all, blank lines are probably not enough for this.

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregatorBase.scala
akkomar: Note to self: automatically catch unused imports.

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregatorBase.scala
akkomar: These are not the same - `fileLimit` is used for reading data via `Dataset` api (I haven't used this, but I can imagine it being useful for testing), whereas `numParquetFiles` controls number of files written per partition (I've clarified it's description).

akkomar:
@fbertsch thanks for comments, RFAL.

src/main/scala/com/mozilla/telemetry/streaming/ExperimentEnrollmentsAggregator.scala
```diff
@@ -15,12 +12,12 @@ import org.apache.spark
 import org.apache.spark.sql.{DataFrame, Dataset, SparkSession}
 import org.rogach.scallop.{ScallopConf, ScallopOption}
```
fbertsch: `ScallopConf` looks to be another unused import

PR #142, opened 2018-06-14
Handle docType parse errors in experiments enrollment aggregator

akkomar:
This introduces a temporary fix for parse error that occurred in production.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/streaming/ExperimentEnrollmentsAggregator.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/ExperimentEnrollmentsAggregator.scala
@@ -99,23 +99,25 @@ object ExperimentEnrollmentsAggregator extends StreamingJobBase {
     import spark.sql.functions._
 
     val events: Dataset[ExperimentEnrollmentEvent] = messages.flatMap(v => {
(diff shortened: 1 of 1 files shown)


src/main/scala/com/mozilla/telemetry/streaming/ExperimentEnrollmentsAggregator.scala
```diff
@@ -99,23 +99,25 @@ object ExperimentEnrollmentsAggregator extends StreamingJobBase {
...
-      val fields = m.fieldsAsMap
-      val docType = fields.getOrElse("docType", "").asInstanceOf[String]
-      if (!allowedDocTypes.contains(docType)) {
-        Array.empty[ExperimentEnrollmentEvent]
-      } else {
-        try {
+      try {
+        val m = Message.parseFrom(v.get(0).asInstanceOf[Array[Byte]])
+        val fields = m.fieldsAsMap
+        // TODO: looks like we can get a double here
```
sunahsuh: What's the actual TODO? investigate this more?

akkomar: Yes, it's for further investigation.

PR #143, opened 2018-06-14
[Bug 1469932] databricks jobs rest api wrapper

haroldwoo:


databricks-jobs/dbjobs.py
```diff
@@ -0,0 +1,213 @@
...
+
+    if (len(job_ids) == 1):
+        return job_ids[0]
+    elif (len(job_ids) < 1):
+        print('GetJobId didnt find any jobs matching env: {}, jarname: {}, version: {}'.format(env, jarname, version))
+        return 0
+    else:
+        print('GetJobId found more than 1 jobs matching env: {}, jarname: {}, version: {}'.format(env, jarname, version))
+        print "Jobs found were: " + str(job_ids)
+        return -1
```
jasonthomas: And -1 here?

databricks-jobs/dbjobs.py
```diff
@@ -0,0 +1,213 @@
...
+                )
+            print('Run_id {} is in a terminal state'.format(runid))
+        except TimeoutException:
+            print('Timed out waiting for run_id {} to terminate'.format(runid))
+
+
+def poll_job_completed(request_string):
+    """ Helper function for polling until the job is in a terminal state """
+    response_json = runs_rpc('get', request_string)
+    if (response_json['state']['life_cycle_state'] == "TERMINATED" or
```
jasonthomas: Maybe instead do something like
```
completed_state = ['TERMINATED', 'SKIPPED', 'INTERNAL_ERROR']

if response_json['state']['life_cycle_state'] in completed_state:
...
```

databricks-jobs/dbjobs.py
```diff
@@ -0,0 +1,213 @@
...
+        response_json['state']['life_cycle_state'] == "SKIPPED" or
+        response_json['state']['life_cycle_state'] == "INTERNAL_ERROR"):
```
jasonthomas: Is `INTERNAL_ERROR` a state we should worry about?

databricks-jobs/sample.py
```diff
@@ -0,0 +1,27 @@
...
+job_id = dbjobs.create_job(JOB_CONF)
+print 'job_id created is ' + str(job_id)
+
+# Update an existing job from json config file
+dbjobs.reset_config(job_id, JOB_CONF)
+
+# Start and stop an existing job
+test_job_id = <YOUR JOB ID>
+run_id = dbjobs.start_job(test_job_id)
+print 'started job ' + str(test_job_id) + ', run id is ' + str(run_id)
```
jasonthomas: str.format() is used above, can we use that here as well?

databricks-jobs/dbjobs.py
```diff
@@ -0,0 +1,213 @@
...
+    for run in response_json['runs']:
+        runids.append(run['run_id'])
+
+    for runid in runids:
+        print('Asynchronously cancelling run_id {} for job_id {}'.format(runid, job_id))
+        request = {}
+        request['run_id'] = runid
+        request_string = json.dumps(request)
+
+        response = requests.post(
```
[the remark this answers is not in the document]
haroldwoo: unfortunately we cant because this api call requires a GET, i can create a new function for it but it was only used once

databricks-jobs/dbjobs.py
```diff
@@ -0,0 +1,213 @@
...
+        jarname: The custom tag "Jar" defined in the job json
+        version: The custom tag "DatasetVersion" defined in the job json
+
+    Returns:
+        job_id: The id of the job found.  0 if no job found, and -1 if more than one job found.
+
+    """
+    job_ids = []
+    response = requests.get(
+        JOBS_BASE_URL + 'list',
```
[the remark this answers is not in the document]
haroldwoo: same as above

databricks-jobs/dbjobs.py
```diff
@@ -0,0 +1,213 @@
...
+                        if tags['Env'] == env and tags['DatasetVersion'] == version and tags['Jar'] == jarname and tags['TelemetryJobName'] == jobname:
+                            job_ids.append(job['job_id'])
+        else:
+            continue
+
+    if (len(job_ids) == 1):
+        return job_ids[0]
+    elif (len(job_ids) < 1):
+        print('GetJobId didnt find any jobs matching env: {}, jarname: {}, version: {}'.format(env, jarname, version))
+        return 0
```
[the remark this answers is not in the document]
haroldwoo: Eventually this script will be used by a jenkins job to do auto deploys on every commit to master.  So i wanted some way of indicating what the issue was without the job having to parse a list

databricks-jobs/dbjobs.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python
+
+import base64
+import json
+import os
+from polling import TimeoutException, poll
+import requests
+
+DOMAIN = 'dbc-caf9527b-e073.cloud.databricks.com'
```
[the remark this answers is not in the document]
haroldwoo: I just never expected it to change

databricks-jobs/dbjobs.py
haroldwoo: nice, thanks

databricks-jobs/dbjobs.py
haroldwoo: not when attempting to stop a job.  If a job errors out we should know beforehand either by email, or build additional monitoring of state.

databricks-jobs/dbjobs.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+def stop_job(job_id):
+    """ Stops all runs of a job, polls until stopped
+
+    Args:
+        job_id: The id of the job to stop all runs for
+
+    """
+    data = {}
```
jasonthomas: You could probably just construct the dict from the beginning since most of these values are already available:

```
data = {
  'active_only' : True
   'job_id': job_id,
   'limit': 0
}

haroldwoo:
good point.  I'm closing this PR and moving this to the cloudops-deployment pr 
https://github.com/mozilla-services/cloudops-deployment/pull/2128/files

PR #146, opened 2018-06-21
Streaming to influx

Ben-Wu:
Sends crashes as string to an http server

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/sinks/BatchHttpSink.scala
+++ b/src/main/scala/com/mozilla/telemetry/sinks/BatchHttpSink.scala
@@ -1,0 +1,93 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package com.mozilla.telemetry.sinks
+
+import org.apache.spark.sql.ForeachWriter
+import scalaj.http.{Http, HttpRequest}
+
+import scala.annotation.tailrec
+import scala.collection.mutable.ArrayBuffer
+import scala.util.control.NonFatal
+import scala.util.{Failure, Success, Try}
+
+class BatchHttpSink(url: String, maxAttempts: Int = 5, defaultDelay: Int = 500, connectionTimeout: Int = 2000,
+                    prefix: String = "", sep: String = "\n", suffix: String = "", maxBatchSize: Int = 1,
+                    retryCodes: List[Int] = List.empty[Int])
+  extends ForeachWriter[String] {
+
+  val TimeoutPseudoCode = -1
+  val ErrorPseudoCode = -2
+
+  val RetryCodes = TimeoutPseudoCode :: Nil ++ retryCodes
+  val OK = 200
+
+  // timeouts in ms
+  val ReadTimeout = 5000
+
+  @transient lazy val log = org.apache.log4j.LogManager.getLogger(this.getClass.getName)
+
+  private val batchCalls = ArrayBuffer[String]()
+
+  override def close(errorOrNull: Throwable): Unit = {
+    flush()
+
+    errorOrNull match {
+      case null =>
+      case e => log.error(e.getStackTrace.mkString("\n"))
+    }
+  }
+
+  override def open(partitionId: Long, version: Long): Boolean = {
+    batchCalls.clear()
+    true
+  }
+
+  override def process(data: String): Unit = {
+    batchCalls.append(data)
+    if (batchCalls.length >= maxBatchSize) {
+      flush()
+    }
+  }
+
+  def flush(): Unit = {
+    if (batchCalls.nonEmpty) {
+      val payload = batchCalls.mkString(prefix, sep, suffix)
+
+      batchCalls.clear()
+
+      attempt(
+        Http(url.toString)
+          .postData(payload)
+          .timeout(connTimeoutMs = connectionTimeout, readTimeoutMs = ReadTimeout))
+    }
+  }
+
+  private def backoff(tries: Int): Long = (scala.math.pow(2, tries) - 1).toLong * defaultDelay
+
+  @tailrec
+  private def attempt(request: HttpRequest, tries: Int = 0): Unit = {
+    if(tries > 0) { // minor optimization
+      java.lang.Thread.sleep(backoff(tries))
+    }
+
+    val code = Try(request.asString.code) match {
+      case Success(c) => c
+      case Failure(e: java.net.SocketTimeoutException) => TimeoutPseudoCode
+      case Failure(e) if NonFatal(e) => {
+        log.error(e.getStackTrace.mkString("\n"))
+        ErrorPseudoCode
+      }
+    }
+
+    (code, tries + 1 == maxAttempts) match {
+      case (OK, _) =>
+      case (ErrorPseudoCode, _) =>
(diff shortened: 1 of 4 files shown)


src/main/scala/com/mozilla/telemetry/streaming/CrashesToInflux.scala
```diff
@@ -0,0 +1,188 @@
...
+      .option("kafka.max.partition.fetch.bytes", 8 * 1024 * 1024) // 8MB
+      .option("spark.streaming.kafka.consumer.cache.maxCapacity", kafkaCacheMaxCapacity)
+      .option("subscribe", TelemetryKafkaTopic)
+      .option("startingOffsets", opts.startingOffsets())
+      .load()
+
+    getParsedPings(pings.select("value"), opts.raiseOnError())
+      .writeStream
+      .queryName(QueryName)
+      .foreach(httpSink)
```
akkomar: As you mentioned, batching these events might be a good idea.
Unfortunetely we can't use foreachPartition on streaming dataset, so this would have to be done in the HttpSink, if needed.

ldct:
lgtm

src/main/scala/com/mozilla/telemetry/streaming/CrashesToInflux.scala
```diff
@@ -0,0 +1,192 @@
...
+
+    datesBetween(opts.from(), opts.to.get).foreach { currentDate =>
+      val dataset = com.mozilla.telemetry.heka.Dataset("telemetry")
+      val pings = dataset
+        .where("sourceName") {
+          case "telemetry" => true
+        }.where("docType") {
+          case doctype if doctype == "crash" => true
+        }.where("appUpdateChannel") { // TODO: make channel filtering work
+          case appUpdateChannel if List[String]("beta").contains(appUpdateChannel) => true
```
Ben-Wu: Using `if opts.acceptedChannels.contains(appUpdateChannel)` gives 
```
org.apache.spark.SparkException: Task not serializable
...
Cause: java.io.NotSerializableException: com.mozilla.telemetry.streaming.CrashesToInflux$Opts
```
Any ideas how to deal with this?

src/main/scala/com/mozilla/telemetry/sinks/RawHttpSink.scala
```diff
@@ -0,0 +1,51 @@
...
+package com.mozilla.telemetry.sinks
+
+import com.mozilla.telemetry.streaming.sinks.HttpSink
+import scalaj.http.{Http, HttpRequest}
+
+import scala.annotation.tailrec
+import scala.util.control.NonFatal
+import scala.util.{Failure, Success, Try}
+
+// TODO: Remove and use generalized HttpSink instead
```
Ben-Wu: I saw that https://github.com/mozilla/telemetry-streaming/pull/147 has an updated HttpSink so this RawHttpSink class wouldn't be needed after that is merged

src/main/scala/com/mozilla/telemetry/streaming/CrashesToInflux.scala
```diff
@@ -0,0 +1,192 @@
...
+      .option("spark.streaming.kafka.consumer.cache.maxCapacity", kafkaCacheMaxCapacity)
+      .option("subscribe", TelemetryKafkaTopic)
+      .option("startingOffsets", opts.startingOffsets())
+      .load()
+
+    getParsedPings(pings.select("value"), opts.raiseOnError(),
+      opts.measurementName(), opts.acceptedChannels(), opts.acceptedAppNames())
+      .writeStream
+      .queryName(QueryName)
+      .foreach(httpSink)
```
Ben-Wu: I still haven't figured out how to properly batch the calls together in streaming.  Any help would be appreciated

akkomar: I think we can handle that in ForeachWriter (copy current implementation of HttpSink: https://github.com/mozilla/telemetry-streaming/blob/master/src/main/scala/com/mozilla/telemetry/sinks/HTTPSink.scala#L30-L44) by:
    1) allocating empty array in `open`
    2) appending events there in `process`
    3) sending collected events in `close`

Another option is implementing `org.apache.spark.sql.execution.streaming.Sink`, which might offer more robust deduplication, but it's much more complicated and I think we don't need to overoptimize on this at this stage.

weivicky:
Good job, Ben!

src/main/scala/com/mozilla/telemetry/streaming/CrashesToInflux.scala
```diff
@@ -0,0 +1,209 @@
...
+      default = Some(1)
+    )
+
+    conflicts(kafkaBroker, List(from, to, fileLimit, maxParallelRequests))
+
+    verify()
+  }
+
+  def parsePing(message: Message, channels: List[String], appNames: List[String],
+                measurementName: String): Array[String] = {
```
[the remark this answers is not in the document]
Ben-Wu: Cool I didn't know that `flatMap` worked like that with Options

src/test/scala/com/mozilla/telemetry/streaming/TestCrashesToInflux.scala
```diff
@@ -0,0 +1,141 @@
...
+import com.github.tomakehurst.wiremock.WireMockServer
+import com.github.tomakehurst.wiremock.client.WireMock
+import com.github.tomakehurst.wiremock.client.WireMock._
+import com.github.tomakehurst.wiremock.core.WireMockConfiguration.wireMockConfig
+import com.holdenkarau.spark.testing.DataFrameSuiteBase
+import com.mozilla.telemetry.sinks.BatchHttpSink
+import org.apache.spark.sql.streaming.StreamingQueryListener
+import org.scalatest.{BeforeAndAfterEach, FlatSpec, Matchers}
+
+class TestCrashesToInflux extends FlatSpec with Matchers with BeforeAndAfterEach with DataFrameSuiteBase {
```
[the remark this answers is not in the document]
Ben-Wu: Should `TestBatchHTTPSink` be renamed too?

akkomar:
Yes, please. I pushed an update today to make this project consistent with
tbv:
https://github.com/mozilla/telemetry-streaming/commit/8e7097fe31e703e4d22d1a16e61b24c55a49baa6

src/main/scala/com/mozilla/telemetry/sinks/BatchHttpSink.scala
```diff
@@ -0,0 +1,92 @@
...
+import scala.util.{Failure, Success, Try}
+
+class BatchHttpSink(url: String, maxAttempts: Int = 5, defaultDelay: Int = 500, connectionTimeout: Int = 2000,
+                    prefix: String = "", sep: String = "\n", suffix: String = "", maxBatchSize: Int = 1)
+  extends ForeachWriter[String] {
+
+  val TimeoutPseudoCode = -1
+  val ErrorPseudoCode = -2
+
+  val RetryCodes = TimeoutPseudoCode :: 429 :: 500 :: 502 :: 503 :: 504 :: Nil
```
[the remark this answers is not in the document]
Ben-Wu: Made the constructor take a list of retry codes instead of using these

Ben-Wu:
@akkomar Ready for rereview

Ben-Wu:
Tests were super flaky and failing inconsistently especially on circleCI.  I just moved `wireMockServer` start/stop to before/after all instead of before/after each and it hasn't failed in the few builds I've tried so far

PR #149, opened 2018-06-30
Event ping events in events dataset

sunahsuh:
The Events ping landed a few days ago in nightly, so this job writes the events coming in over the event pings into the `events` dataset (which is already partitioned by docType, how convenient!) This will be paired with an update to the main ping events job in telemetry-batch-view that updates the schema to be in line with the one laid out here (but the schema changes should be backwards-compatible so I'm opting not to bump the version.)

I plan to schedule this as a batch job for now but the job ended up here instead of in telemetry-batch-view since what I'm aiming for in Q3 is to create a streaming job that takes in both main and event pings, explodes out the events and publishes them in to a new kafka topic to be consumed by both the job that writes out the events dataset as well as jobs that send to amplitude, as well as any aggregations to be done on realtime events (e.g. experiment enrollments, etc)

One note: I tried to use the dynamic partition overwrite config in my tests and it wouldn't work -- there's something wonky going on with the write committer that I'd like to discuss when you're back (the spark logs say it worked correctly including the copy from the temp directory but the new partitions aren't there when I look.)

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/streaming/EventPingEvents.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/EventPingEvents.scala
@@ -1,0 +1,141 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package com.mozilla.telemetry.streaming
+
+import com.mozilla.telemetry.heka.{Message, Dataset => MozDataset}
+import com.mozilla.telemetry.pings.EventPing
+import com.mozilla.telemetry.streaming.StreamingJobBase.TelemetryKafkaTopic
+import org.apache.spark.sql.{DataFrame, Dataset, SparkSession}
+import org.rogach.scallop.ScallopOption
+
+object EventPingEvents extends StreamingJobBase {
+  override val outputPrefix = "events/v1"
+
+  val kafkaCacheMaxCapacity = 10
+
+  private val allowedDocTypes = List("event")
+  private val allowedAppNames = List("Firefox")
+
+  private class Opts(args: Array[String]) extends BaseOpts(args) {
+    val outputPath: ScallopOption[String] = opt[String](
+      "outputPath",
+      descr = "Output path",
+      required = true)
+    val failOnDataLoss: ScallopOption[Boolean] = opt[Boolean](
+      "failOnDataLoss",
+      descr = "Whether to fail the query when it’s possible that data is lost.")
+    val maxRecordsPerFile = opt[Int]("max-records-per-file",
+      descr = "Max number of rows to write to output files before splitting",
+      required = false, default = Some(10000000))
+
+    conflicts(kafkaBroker, List(from, to, maxRecordsPerFile))
+    verify()
+  }
+
+  def main(args: Array[String]): Unit = {
+    val opts = new Opts(args)
+
+    val spark = SparkSession.builder()
+      .appName("Event Ping Events")
+      .config("spark.streaming.stopGracefullyOnShutdown", "true")
+      .getOrCreate()
+
+    opts.kafkaBroker.get match {
+      case Some(_) => writeStreamingEvents(spark, opts)
+      case None => writeBatchEvents(spark, opts)
+    }
+  }
+
+  def writeStreamingEvents(spark: SparkSession, opts: EventPingEvents.Opts): Unit = {
+    val pings = spark
+      .readStream
+      .format("kafka")
+      .option("kafka.bootstrap.servers", opts.kafkaBroker())
+      .option("failOnDataLoss", opts.failOnDataLoss())
+      .option("kafka.max.partition.fetch.bytes", 8 * 1024 * 1024) // 8MB
+      .option("spark.streaming.kafka.consumer.cache.maxCapacity", kafkaCacheMaxCapacity)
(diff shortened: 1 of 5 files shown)


src/main/scala/com/mozilla/telemetry/streaming/EventPingEvents.scala
```diff
@@ -0,0 +1,149 @@
...
+      logger.info(s"Event ping events written for $currentDate")
+    }
+
+    spark.stop()
+  }
+
+  private[streaming] def explodeEvents(messages: DataFrame): Dataset[EventRow] = {
+    import messages.sparkSession.implicits._
+
+    messages.flatMap(v => {
```
jklukas: I'm used to seeing this as:

    messages.flatMap { v =>

which avoids an extra set of parentheses but I believe is otherwise identical.

src/main/scala/com/mozilla/telemetry/streaming/EventPingEvents.scala
```diff
@@ -0,0 +1,149 @@
...
+                ping.payload.sessionId, ping.payload.subsessionId, ping.sessionStart,
+                (ping.meta.Timestamp / 1e9).toLong, ping.meta.sampleId.map(_.toString), ping.getMSStyleExperiments,
+                e.timestamp, e.category, e.method, e.`object`, e.value, e.extra, process)
+
+            }
+          }
+        }
+      /*
+      } catch {
+        // TODO: track parse errors
```
jklukas: What's current behavior if we have parse errors? Does the whole job shut down?

src/main/scala/com/mozilla/telemetry/streaming/EventPingEvents.scala
```diff
@@ -0,0 +1,149 @@
...
+
+    spark.stop()
+  }
+
+  private[streaming] def explodeEvents(messages: DataFrame): Dataset[EventRow] = {
+    import messages.sparkSession.implicits._
+
+    messages.flatMap(v => {
+      //try {
+        val m = Message.parseFrom(v.get(0).asInstanceOf[Array[Byte]])
```
jklukas: Looks like we can use `v.getAs[Array[Byte]](0)` here.

src/main/scala/com/mozilla/telemetry/streaming/EventPingEvents.scala
```diff
@@ -0,0 +1,149 @@
...
+        val fields = m.fieldsAsMap
+        val docType = fields.getOrElse("docType", "").asInstanceOf[String]
+        if (!allowedDocTypes.contains(docType)) {
```
jklukas: I'd prefer to reverse the order of the branches here so we can remove the `!` and avoid a double negative.

src/test/scala/com/mozilla/telemetry/TestUtils.scala
```diff
@@ -367,6 +367,124 @@ object TestUtils {
...
+         |        [
+         |          4203542,
+         |          "test",
+         |          "no",
+         |          "extras"
+         |        ]
+         |      ]
+         |    }
+       """.stripMargin
+    1.to(size) map { index =>
```
jklukas: I'm used to seeing this as `(1 to size).map`

src/main/scala/com/mozilla/telemetry/streaming/EventPingEvents.scala
sunahsuh: That's some copy-pasta from using the streaming experiment enrollments job as a base :) Because this job can run as a streaming job as well as in batch mode, we don't want to throw on parse errors and kill the job but also we probably start tracking errors and logging them instead of eating them silently

src/main/scala/com/mozilla/telemetry/streaming/StreamingJobBase.scala
```diff
@@ -28,6 +28,8 @@ abstract class StreamingJobBase extends Serializable {
 
   val clock: Clock = Clock.systemUTC()
 
+  private[streaming] val logger = org.apache.log4j.Logger.getLogger(this.getClass.getName)
```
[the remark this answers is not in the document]
sunahsuh: That *does* seem like a nifty solution, and I'm totally in favor. Unfortunately, naively switching to logback-classic for the backend spewed out a bunch of logs into the test output, and I don't really have the time to look into what's happening with that right now. I just removed all the logging from the job and I'll file a bug re: using scala-logging.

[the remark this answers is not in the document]
jklukas: I'm totally fine with that.

src/main/scala/com/mozilla/telemetry/streaming/EventPingEvents.scala
sunahsuh: We have a few instances of this pattern in the codebase already -- I don't have a preference except a slight one for staying consistent

jklukas: Consistency is a great reason for keeping this as-is.

src/main/scala/com/mozilla/telemetry/streaming/StreamingJobBase.scala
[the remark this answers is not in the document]
akkomar: @sunahsuh @jklukas be careful with that - iirc Spark uses both slf4j and directly log4j, so switching backends will be tricky.

PR #154, opened 2018-07-17
Bug 1462107 - Add frecency aggregator

akkomar:


src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
```diff
@@ -0,0 +1,151 @@
...
+import org.apache.spark.deploy.SparkHadoopUtil
+import org.apache.spark.sql.execution.streaming.Sink
+import org.apache.spark.sql.sources.StreamSinkProvider
+import org.apache.spark.sql.streaming.OutputMode
+import org.apache.spark.sql.{DataFrame, SQLContext}
+import org.json4s.DefaultFormats
+import org.json4s.jackson.Serialization
+
+object FederatedLearningSearchOptimizerConstants {
+  // https://github.com/florian/federated-learning-addon/blob/master/addon/prefs.js#L26
```
florian: The default values are not in the addon anymore, so that link doesn't work anymore.
If you want a permalink to the current default values: https://dxr.mozilla.org/mozilla-central/rev/085cdfb90903d4985f0de1dc7786522d9fb45596/browser/app/profile/firefox.js#901

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
```diff
@@ -0,0 +1,151 @@
...
+        val lastWeights: Array[Double] = state.weights
+        val learningRates: Array[Double] = state.learningRates
+        val previousGradient: Option[Array[Double]] = state.gradient
+
+        val gradient = aggregate.avgUpdates
+        val InternalResult(newWeights, newLearningRates) = FederatedLearningRPropOptimizer.fit(lastWeights, gradient, previousGradient, learningRates)
+
+        val newIteration = iteration + 1
+
+        // val log = Array(OptimisationLogLine(aggregate.window.start, aggregate.modelVersion.toLong, aggregate.avgLoss))
```
florian: That should probably be removed before merging.

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
```diff
@@ -0,0 +1,151 @@
...
+
+        val newState = OptimisationState(newIteration, newWeights, newLearningRates, Option(gradient))
+        writeModel(ModelOutput(newWeights.map(math.round(_).toInt), newIteration))
+
+        writeState(newState)
+        state = newState
+    }
+  }
+
+  private[federated] def writeModel(modelOutput: ModelOutput): Unit = {
```
florian: I'm not sure if you're already giving the file to s3 here, but two points about s3:
- The file permissions need to be changed so that the file is accessible for everyone
- Caching should be disabled (HTTP header `Cache-Control: no-cache` so that clients always fetch the newest model)

Anthony did both of these for me so that I could test my addon, but he needed to explicitly configure both of them.

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
```diff
@@ -0,0 +1,151 @@
...
+    implicit val formats = DefaultFormats
+    val jsonModel = Serialization.write(modelOutput)
```
florian: Just to make sure: This creates a JSON object with two keys `model` and `iteration`, like in [this example](https://s3-us-west-2.amazonaws.com/telemetry-test-bucket/frecency/latest.json)?

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
```diff
@@ -0,0 +1,151 @@
...
+
+  private[federated] var state: OptimisationState = initState()
+
+  override def addBatch(batchId: Long, data: DataFrame): Unit = {
+    val aggregates = data.collect().map(FrecencyUpdateAggregate(_))
+
+    val iteration: Long = state.iteration
+
+    val ord = Ordering.by((_: FrecencyUpdateAggregate).modelVersion)
+    aggregates.filter(_.studyVariation == "treatment").filter(_.modelVersion.toLong >= iteration).reduceOption(ord.min) match {
```
sunahsuh: Isn't this, pragmatically, equivalent to filtering by `_.modelVersion == iteration`? I'm also curious why you chose `>=` -- when would you expect to see a model version that's greater than the current iteration?

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
```diff
@@ -0,0 +1,151 @@
...
+      case Some(aggregate) =>
+
+        val lastWeights: Array[Double] = state.weights
+        val learningRates: Array[Double] = state.learningRates
+        val previousGradient: Option[Array[Double]] = state.gradient
+
+        val gradient = aggregate.avgUpdates
+        val InternalResult(newWeights, newLearningRates) = FederatedLearningRPropOptimizer.fit(lastWeights, gradient, previousGradient, learningRates)
+
+        val newIteration = iteration + 1
```
sunahsuh: I had envisioned that the iterations would be named via the current timestamp -- I can see the advantage here too (knowing how many iterations we've run at a glance.)

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
```diff
@@ -0,0 +1,151 @@
...
+    outputStream.close()
+    fs.close()
+  }
+
+  private[federated] def initState(): OptimisationState = {
+    val conf = SparkHadoopUtil.get.conf
+    val checkpointPath = new Path(stateCheckpointPath)
+    val fs = checkpointPath.getFileSystem(conf)
+    if (!fs.exists(checkpointPath)) {
+      fs.mkdirs(checkpointPath)
```
sunahsuh: Hmm, how does this work for S3, re: keys not existing if there's no file inside

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
```diff
@@ -0,0 +1,151 @@
...
+      latestFileOpt match {
+        case Some(latestFile) =>
+          val fis = fs.open(latestFile)
+          val fileContents: String = IOUtils.toString(fis)
+          fis.close()
+
+          implicit val formats = DefaultFormats
+          Serialization.read[OptimisationState](fileContents)
+        case None =>
+          OptimisationState(0, StartingWeights, Array.fill(NumberOfFeatures)(StartingLearningRate), None)
```
sunahsuh: Can we add a config option that points to a file we can read in for bootstrapping OptimisationState? I think that might come in really handy if we need to restart from a particular point

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
sunahsuh: It seems likely we'll have to switch to an S3-specific library for these requirements

sunahsuh:
Oh wait, note the failing test (which doesn't seem related, bleh)

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
akkomar: Yes, like here: https://github.com/mozilla/telemetry-streaming/blob/frecency_aggregator/src/test/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3SinkTest.scala#L56

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
akkomar: I'll just log this and we'll aggregate logs from the cluster to s3 for inspection.

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
akkomar: Good idea, adding this option.

src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
```diff
@@ -0,0 +1,128 @@
...
+      try {
+        val m = Message.parseFrom(v.get(0).asInstanceOf[Array[Byte]])
+        val fields = m.fieldsAsMap
+        val docType = fields.getOrElse("docType", "").asInstanceOf[String]
+        if ("frecency-update" == docType) {
+          val ping = FrecencyUpdatePing(m)
+          Option(FrecencyUpdate(
+            ping.meta.normalizedTimestamp(),
+            ping.payload.model_version,
+            ping.payload.study_variation,
```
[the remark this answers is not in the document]
akkomar: Yeah, you're right, looks like it's a leftover from prototyping.

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
akkomar: Good question - I don't know when this might occur (most likely when we misconfigure something).
I think I wanted to be more careful by not dropping too much data, but when I think of it now I'm not sure if it makes much difference.
I'll think about it, let me know if you think it better to switch to equality comparison here.

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
akkomar: Oh, I was used to this convention, we also agreed with @florian to use iteration number when naming model outputs. We can add timestamps to the number if needed.

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
akkomar: It works :)

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
florian: Cool, that looks good 👍

src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
```diff
@@ -0,0 +1,152 @@
...
+
+import scala.collection.mutable
+
+object FederatedLearningSearchOptimizer extends StreamingJobBase {
+
+  def main(args: Array[String]): Unit = {
+    val opts = new Opts(args)
+
+    val spark = SparkSession.builder()
+      .appName("Experiment Enrollments to TestTube")
```
sunahsuh: Copy-pasta

src/test/scala/com/mozilla/telemetry/learning/federated/FederatedLearningRPropOptimizerTest.scala
```diff
@@ -0,0 +1,56 @@
...
+    optimizedWeights.indices.foreach { i =>
+      val expected = weights(i + 1)
+      val calculated = optimizedWeights(i)
+
+      expected.zip(calculated).foreach { case (e, c) =>
+        assert(e === c)
+      }
+
+      // scalastyle:off print-ln
+      println(s"Iteration $i")
```
sunahsuh: Do you want to keep these printlns in the test output?

src/test/scala/com/mozilla/telemetry/util/ManualClock.scala
```diff
@@ -0,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package com.mozilla.telemetry.util
+
+import java.io.Serializable
+import java.time.{Clock, Instant, ZoneId}
+
+class ManualClock(var instant: Instant, val zone: ZoneId) extends Clock with Serializable {
```
sunahsuh: This is super cool :D

src/test/scala/com/mozilla/telemetry/learning/federated/FederatedLearningRPropOptimizerTest.scala
akkomar: I'm removing printlns and will log only in case of errors.

src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningRPropOptimizer.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  private def applyConstraints(weights: Array[Double], alignTimeBuckets: Boolean): Array[Double] = {
+    log.info(s"Weights before constraints: ${weights.mkString(",")}")
+    val nonNegativeWeights = (weights.min match {
+      case min if min < 0 => weights.map(_ - min)
+      case _ => weights
+    }).toBuffer
+
+    if (alignTimeBuckets) {
+      nonNegativeWeights(1) = math.max(nonNegativeWeights(1), nonNegativeWeights(0) + 1)
```
florian: @akkomar @sunahsuh I think we should add one more constraint to the start of the if:
`nonNegativeWeights(0) = math.max(nonNegativeWeights(0), 1)`

This would be to make sure that the first time bucket always captures at least one day.

PR #155, opened 2018-07-18
Bug 1476789 - Add channel option to ErrorAggregator

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1476789)

The `experiments_error_aggregates` job in airflow is a dependency of the `experiments_aggregates` job in the `main_summary` DAG. I'm currently setting the channel to nightly in the MainSummaryView when the deploy environment is `dev`. 

This PR exposes an option to set a channel to filter in the options. I also set the `minPartitions` option in `Dataset.records` so that the job finishes in a reasonable amount of time on a single machine. I think this should work out on the current production settings.

In order to avoid spilling, the partitions should be roughly this size: 
`size(partition) <= memory * spark.memory.fraction / spark.default.parallelism`
or roughly `~30GB * 0.6 / 32 <= 562.5 MB` with c3.x4large nodes.

With our current settings, we should be able to fit around 1.4TB of data without spilling into disk on batch mode.

`562.5MB * 20 nodes * 32 executors * 4 = 1.44TB >= size(Dataset)`

I've tested that the current behavior is unchanged and that the new behavior is suitable for testing.

```
date="20180701"

spark-submit --master yarn \
             --deploy-mode client \
             --class com.mozilla.telemetry.streaming.ExperimentsErrorAggregator \
             telemetry-streaming-assembly-0.1-SNAPSHOT.jar \
             --outputPath "s3://telemetry-test-bucket" \
             --numParquetFiles 6 \
             --from $date \
             --to $date
             --channel nightly

aws s3 ls s3://telemetry-test-bucket/experiment_error_aggregates/v1/submission_date_s3=20180701/ --summarize --human-readable
2018-07-18 15:29:33  200.5 KiB part-00000-ef15f220-7199-4aad-92f6-effbff17e187.c000.snappy.parquet
2018-07-18 15:29:33  201.8 KiB part-00001-ef15f220-7199-4aad-92f6-effbff17e187.c000.snappy.parquet
2018-07-18 15:29:33  201.7 KiB part-00002-ef15f220-7199-4aad-92f6-effbff17e187.c000.snappy.parquet
2018-07-18 15:29:33  202.1 KiB part-00003-ef15f220-7199-4aad-92f6-effbff17e187.c000.snappy.parquet
2018-07-18 15:29:33  202.6 KiB part-00004-ef15f220-7199-4aad-92f6-effbff17e187.c000.snappy.parquet
2018-07-18 15:29:33  200.2 KiB part-00005-ef15f220-7199-4aad-92f6-effbff17e187.c000.snappy.parquet

Total Objects: 6
   Total Size: 1.2 MiB
```

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
@@ -52,6 +52,10 @@ object ErrorAggregator {
       "fileLimit",
       descr = "Max number of files to retrieve (batch mode only). Default: All files",
       required = false)
+    val channel: ScallopOption[String] = opt[String](
+      "channel",
+      descr = "Only process data from the given channel",
+      required = false)
     val outputPath: ScallopOption[String] = opt[String](
       "outputPath",
       descr = "Output path",
@@ -322,6 +326,7 @@ object ErrorAggregator {
 
     for (offset <- 0 to Days.daysBetween(from, to).getDays) {
       val currentDate = from.plusDays(offset)
+      val filterChannel = opts.channel.get
 
       val pings = Dataset("telemetry")
         .where("sourceName") {
@@ -332,7 +337,9 @@ object ErrorAggregator {
           case appName if allowedAppNames.contains(appName) => true
         }.where("submissionDate") {
           case date if date == currentDate.toString(dateFormat) => true
-        }.records(opts.fileLimit.get)
+        }.where("appUpdateChannel") {
+          case channel => filterChannel.isEmpty || channel == filterChannel.get
+        }.records(opts.fileLimit.get, Some(sc.defaultParallelism * 4))
         .map(m => Row(m.toByteArray))
 
       val schema = StructType(List(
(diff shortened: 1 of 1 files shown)


acmiyaguchi:
I was thinking of this option as a batch only mode, but it might be useful for debugging the streaming mode too. Is it possible to run the current streaming job on a single node against any of our queues, @fbertsch?

fbertsch:
We've only been running on databricks as of late, but we've definitely ran it on ATMO in the past. The machines do have access to (both, I think) kafka instances, so we just need the URL. @akkomar or @haroldwoo  would be the people to ask about that.

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
```diff
@@ -52,6 +52,10 @@ object ErrorAggregator {
       "fileLimit",
       descr = "Max number of files to retrieve (batch mode only). Default: All files",
       required = false)
+    val channel: ScallopOption[String] = opt[String](
```
fbertsch: This should conflict with `kafkaBroker`, on line 88. Add that and we're good to go!

acmiyaguchi:
I opted against making more changes and just documented that the option is batch mode specific. It looks like there's a refactor in flight, and it's a bit out of the scope of the end-to-end airflow tests.

src/main/scala/com/mozilla/telemetry/streaming/ErrorAggregator.scala
acmiyaguchi: Right on, added it to the `conflicts` list in an amended commit.

PR #156, opened 2018-07-19
Add logs to federated learning optimizer

sunahsuh:
I think this hits all the logs we'd want (besides the per-branch monitors, which we'll do in hindsight)


src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala
```diff
@@ -29,6 +29,8 @@ class FederatedLearningSearchOptimizerS3Sink(outputPath: String, stateCheckpoint
   private[federated] var state: OptimisationState = initState()
 
   override def addBatch(batchId: Long, data: DataFrame): Unit = {
+    log.info(s"Adding new batch with starting state $state")
```
akkomar: I'd move this below to the place when we're sure we have some data. Sink is called with empty dataframe many times between window bounds - I think this log might be hard to interpret if left here.

sunahsuh: The intent on my end was to know that the addBatch method was triggered -- I can tweak the wording if it would make it clearer

PR #160, opened 2018-07-26
Bug 1474987 Add pipeline for Savant events to Amplitude

jklukas:
This adds the JSON file to support Savant event mappings and adds a few new features to support those mappings.

First, we now support literal fields. Values enclosed in single quotes will output the literal value rather than looking up the property in the event.

Second, we now support dataset sources other than "telemetry".

### DIFF
--- a/configs/desktop_savant_events_schemas.json
+++ b/configs/desktop_savant_events_schemas.json
@@ -1015,7 +1015,7 @@
                   "pwmgr"
                 ]
               },
-              "value": {
+              "object": {
                 "type": "string",
                 "enum": [
                   "ask"
@@ -1026,7 +1026,7 @@
               "timestamp",
               "category",
               "method",
-              "value"
+              "object"
             ]
           }
         },
@@ -1058,7 +1058,7 @@
                   "pwmgr"
                 ]
               },
-              "value": {
+              "object": {
                 "type": "string",
                 "enum": [
                   "save"
@@ -1069,7 +1069,7 @@
               "timestamp",
               "category",
               "method",
-              "value"
+              "object"
             ]
           }
         },
@@ -1101,7 +1101,7 @@
                   "pwmgr"
                 ]
               },
-              "value": {
+              "object": {
                 "type": "string",
                 "enum": [
                   "update"
@@ -1112,7 +1112,7 @@
               "timestamp",
               "category",
               "method",
-              "value"
+              "object"
             ]
           }
         },
--- a/configs/desktop_savant_events_schemas.json
+++ b/configs/desktop_savant_events_schemas.json
@@ -1,0 +1,1157 @@
+{
+  "source": "telemetry-cohorts",
+  "filters": {
+    "docType": [
+      "main"
+    ],
+    "appName": [
+      "Firefox"
+    ],
+    "experimentId": [
+      "pref-flip-savant-1457226-de-existing-users",
+      "pref-flip-savant-1457226-de-new-users",
+      "pref-flip-savant-1457226-en-existing-users",
+      "pref-flip-savant-1457226-en-new-users"
+    ]
+  },
+  "eventGroups": [
+    {
+      "eventGroupName": "SAVANT",
+      "events": [
+        {
+          "name": "open bookmark",
+          "description": "client opens a bookmark",
+          "amplitudeProperties": {
+            "meta-cat-1": "literal.navigate",
+            "meta-cat-2": "literal.bookmark"
+          },
+          "schema": {
+            "$schema": "http://json-schema.org/schema#",
+            "type": "object",
+            "properties": {
+              "timestamp": {
+                "type": "number",
+                "minimum": 0
+              },
+              "category": {
+                "type": "string",
+                "enum": [
+                  "savant"
+                ]
+              },
+              "method": {
+                "type": "string",
+                "enum": [
+                  "follow_bookmark"
+                ]
+              },
+              "object": {
+                "type": "string",
+                "enum": [
+                  "open"
+                ]
+              }
+            },
+            "required": [
+              "timestamp",
+              "category",
+              "method",
+              "object"
+            ]
+          }
+        },
+        {
+          "name": "save bookmark",
+          "description": "client saves a bookmark",
+          "amplitudeProperties": {
+            "meta-cat-1": "literal.feature",
+            "meta-cat-2": "literal.bookmark"
+          },
+          "schema": {
+            "$schema": "http://json-schema.org/schema#",
+            "type": "object",
+            "properties": {
+              "timestamp": {
+                "type": "number",
+                "minimum": 0
+              },
+              "category": {
+                "type": "string",
+                "enum": [
+                  "savant"
+                ]
+              },
+              "method": {
+                "type": "string",
+                "enum": [
+                  "bookmark"
+                ]
+              },
+              "object": {
+                "type": "string",
+                "enum": [
+                  "save"
+                ]
+              }
+            },
+            "required": [
+              "timestamp",
+              "category",
+              "method",
+              "object"
+            ]
+          }
+        },
+        {
+          "name": "remove bookmark",
+          "description": "client removes a bookmark",
+          "amplitudeProperties": {
+            "meta-cat-1": "literal.feature",
+            "meta-cat-2": "literal.bookmark"
+          },
+          "schema": {
+            "$schema": "http://json-schema.org/schema#",
+            "type": "object",
+            "properties": {
+              "timestamp": {
+                "type": "number",
+                "minimum": 0
+              },
+              "category": {
+                "type": "string",
+                "enum": [
+                  "savant"
+                ]
+              },
+              "method": {
+                "type": "string",
+                "enum": [
+                  "bookmark"
+                ]
+              },
+              "object": {
+                "type": "string",
+                "enum": [
+                  "remove"
+                ]
+              }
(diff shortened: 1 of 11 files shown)


configs/desktop_savant_events_schemas.json
```diff
@@ -0,0 +1,1122 @@
...
+    ],
+    "appName": [
+      "Firefox"
+    ],
+    "experimentId": [
+      "pref-flip-savant-1457226-de-existing-users",
+      "pref-flip-savant-1457226-de-new-users",
+      "pref-flip-savant-1457226-en-existing-users",
+      "pref-flip-savant-1457226-en-new-users"
+    ]
```
jklukas: This file is copied directly from https://github.com/mozilla/SAVANT-study-analysis/blob/98cfa95e0141e12595c08ea96932a95885368e11/savantConfigFile.json

The only modifications I've made are to add the "source" field at the top and added this filter on experimentId.

src/test/scala/com/mozilla/telemetry/streaming/EventsToAmplitudeTest.scala
```diff
@@ -87,8 +87,27 @@ class EventsToAmplitudeTest extends FlatSpec with Matchers with BeforeAndAfterAl
...
     s"""{ "event_type": "$eventGroup - AppOpen" }""" ::
-    s"""{ "event_type": "$eventGroup - Erase", "event_properties": { "erase_object": "erase_home" }, "user_properties": { "host": "side" }}""" ::
-    s"""{ "event_type": "second_event_group - AppClose", "event_properties": { "session_length": "1000" }}""" :: Nil
+    s"""
+      |{
+      |   "user_properties" : {
+      |      "host" : "side"
+      |   },
+      |   "event_properties" : {
+      |      "literal_field" : "literal value",
```
jklukas: This line is the only addition. Otherwise, these strings should be equivalent to the single-line versions modulo whitespace.

configs/desktop_savant_events_schemas.json
```diff
@@ -0,0 +1,1122 @@
+{
+  "source": "telemetry-cohorts",
+  "filters": {
+    "docType": [
+      "main"
+    ],
+    "appName": [
+      "Firefox"
+    ],
```
jklukas: appName is not a partition field for "telemetry-cohorts", but I assume that's fine; this will just get treated as a ping-level filter rather than a partitioning filter.

sunahsuh: IIRC the experiment was release-only -- should we also filter on that?

configs/desktop_savant_events_schemas.json
```diff
@@ -0,0 +1,1122 @@
...
+                "type": "string",
+                "enum": [
+                  "hamburger_menu"
+                ]
+              }
+            },
+            "required": [
+              "timestamp",
+              "category",
+              "method"
```
sunahsuh: According to the [telemetry event definition](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/telemetry/Events.yaml#206) this should also have an object called "open"

configs/desktop_savant_events_schemas.json
```diff
@@ -0,0 +1,1122 @@
...
+                "type": "string",
+                "enum": [
+                  "library_menu"
+                ]
+              }
+            },
+            "required": [
+              "timestamp",
+              "category",
+              "method"
```
sunahsuh: Same kind of thing here, re: the object (I do think it's worth explicitly adding since if events.yaml changes and we add a "close" event then we'd catch both of those)

configs/desktop_savant_events_schemas.json
```diff
@@ -0,0 +1,1122 @@
...
+                  "savant"
+                ]
+              },
+              "method": {
+                "type": "string",
+                "enum": [
+                  "pwmgr"
+                ]
+              },
+              "value": {
```
sunahsuh: [Events.yaml](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/telemetry/Events.yaml#246) suggests this, and the other pwmgr events should be looking at "object" instead of "value"

configs/desktop_savant_events_schemas.json
```diff
@@ -0,0 +1,1122 @@
...
+                "type": "string",
+                "enum": [
+                  "pwmgr_use"
+                ]
+              }
+            },
+            "required": [
+              "timestamp",
+              "category",
+              "method"
```
sunahsuh: Again, re: explicitly specifying object ("use", in this case)

configs/desktop_savant_events_schemas.json
```diff
@@ -0,0 +1,1122 @@
...
+                "type": "string",
+                "enum": [
+                  "dotdotdot_menu"
+                ]
+              }
+            },
+            "required": [
+              "timestamp",
+              "category",
+              "method"
```
sunahsuh: Same re: `object`

configs/desktop_savant_events_schemas.json
```diff
@@ -0,0 +1,1122 @@
...
+                "type": "string",
+                "enum": [
+                  "follow_bookmark"
+                ]
+              }
+            },
+            "required": [
+              "timestamp",
+              "category",
+              "method"
```
sunahsuh: Same, re: object

src/main/scala/com/mozilla/telemetry/pings/Ping.scala
```diff
@@ -389,6 +395,7 @@ case class Event(timestamp: Int,
     case "object" => `object`
     case "value" => value.getOrElse ("")
     case e if e.startsWith ("extra") => extra.getOrElse (Map.empty).getOrElse (e.stripPrefix ("extra."), "")
+    case Event.literalFieldPattern(unquoted) => unquoted
```
sunahsuh: Slight preference to keep the convention above re: "extra" and use "literal.<value>"

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
```diff
@@ -204,10 +218,11 @@ object EventsToAmplitude extends StreamingJobBase {
     implicit val sc = spark.sparkContext
 
     datesBetween(opts.from(), opts.to.get).foreach { currentDate =>
-      val dataset = com.mozilla.telemetry.heka.Dataset("telemetry")
+      val dataset = com.mozilla.telemetry.heka.Dataset(config.source)
+      val topLevelFields = TOP_LEVEL_PING_FIELDS(config.source)
 
       val pings = config.getBatchFilters.filter{
-          case(name, _) => TOP_LEVEL_PING_FIELDS.contains(name)
+          case(name, _) => topLevelFields.contains(name)
```
sunahsuh: Oh, this is kind of silly duplicated code -- let's fix this so topLevelFilters is correctly formatted

configs/desktop_savant_events_schemas.json
jklukas: Su asked in the bug about whether it's necessary to provide the full coordinates, or just enough to uniquely identify the event.

If there are no other events with category=savant and method=hamburger_menu, there's no functional difference here from specifying the object, or am I misunderstanding?

src/main/scala/com/mozilla/telemetry/pings/Ping.scala
jklukas: Fair. Will change.

configs/desktop_savant_events_schemas.json
sunahsuh: There aren't *now*, but the definitions could change under us – I see no disadvantage to being explicit, and an advantage of some safety

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
jklukas: I don't have the context to understand why we need to translate from `-` to `_`. Are there some contexts where we want the dashed form. Is it safe to always do the replacement?

configs/desktop_savant_events_schemas.json
jklukas: I'll ask Su to confirm.

configs/desktop_savant_events_schemas.json
jklukas: More modifications have been made, so the above is no longer really true.

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
sunahsuh: Oh, shoot, you're right -- on disk (e.g. reading in batch pings) the docType is normalized and `-` are replaced with `_`, but in the ping fields it's still a dash

configs/desktop_savant_events_schemas.json
jklukas: The `description`s for those events make reference to what will be populated in the "value field". I don't understand why both object and value exist with similar settings.

SuYoungHong: @sunahsuh , correct, good catch, that's a typo on my part

@jklukas , good point. the difference is when client is asked if they want to save/update a password, we record object as ask, and depending on what was asked, put save or update in the value. when the client chooses to save or update, the event becomes object=save or update, with a blank value. 

definitely worth redesigning in the future

PR #162, opened 2018-07-30
Bug 1474987 Send additional client properties

jklukas:
Adds the fields requested in https://github.com/mozilla/SAVANT-study-analysis/blob/ed5c4c706ad9b5ab0bd0bacc29b0daff7d00920a/AMP_CLIENTS.md as user properties sent to Amplitude.


src/main/scala/com/mozilla/telemetry/pings/Ping.scala
```diff
@@ -367,11 +367,19 @@ trait SendsToAmplitudeWithEnvironment extends SendsToAmplitude {
...
+      ("app_build_id" -> meta.appBuildId) ~
+      ("app_name" -> meta.appName) ~
+      ("app_version" -> meta.appVersion) ~
+      ("locale" -> meta.`environment.settings`.map(_.locale)) ~
+      ("is_default_browser" -> meta.`environment.settings`.map(_.isDefaultBrowser)) ~
+      ("country" -> meta.geoCountry) ~
+      ("os" -> meta.os) ~
+      ("os_version" -> meta.`environment.system`.map(_.os.name)) ~
+      ("env_build_arch" -> meta.`environment.build`.map(_.architecture)) ~
+      ("experiments" -> experimentsArray)
```
jklukas: Starting to fill out the values requested by @SuYoungHong in https://github.com/mozilla/SAVANT-study-analysis/blob/master/AMP_CLIENTS.md

It's not clear to me how to get at properties outside of meta (memoryMB, for example, is contained in system). Will chat with @sunahsuh about approaches.

src/main/scala/com/mozilla/telemetry/pings/Ping.scala
```diff
@@ -367,11 +379,23 @@ trait SendsToAmplitudeWithEnvironment extends SendsToAmplitude {
 
     ("user_properties" ->
       ("channel" -> meta.normalizedChannel) ~
```
jklukas: The only thing missing here is that https://github.com/mozilla/SAVANT-study-analysis/blob/ed5c4c706ad9b5ab0bd0bacc29b0daff7d00920a/AMP_CLIENTS.md mentions both channel and normalized_channel.

We are already including normalized channel as "channel". Is raw channel necessary, @SuYoungHong?

SuYoungHong: @jklukas , good point. I suppose it's fine to exclude it. it's important for historical data, but since this study is on new data, it's fine to exclude the raw channel.

src/main/scala/com/mozilla/telemetry/pings/Ping.scala
```diff
@@ -367,11 +379,23 @@ trait SendsToAmplitudeWithEnvironment extends SendsToAmplitude {
...
-      ("user_id" -> getClientId)
+      ("sample_id" -> meta.sampleId) ~
+      ("app_build_id" -> meta.appBuildId) ~
+      ("app_name" -> meta.appName) ~
+      ("app_version" -> meta.appVersion) ~
+      ("locale" -> meta.`environment.settings`.map(_.locale)) ~
+      ("is_default_browser" -> meta.`environment.settings`.map(_.isDefaultBrowser)) ~
+      ("country" -> meta.geoCountry) ~
+      ("os" -> meta.os) ~
+      ("os_version" -> meta.`environment.system`.map(_.os.name)) ~
```
sunahsuh: app_version, country, os and os_version are all top-level amplitude fields that are already sent over (see lines 310-313)

src/main/scala/com/mozilla/telemetry/pings/Ping.scala
```diff
@@ -367,11 +379,23 @@ trait SendsToAmplitudeWithEnvironment extends SendsToAmplitude {
...
+      ("env_build_arch" -> meta.`environment.build`.map(_.architecture)) ~
+      ("is_wow64" -> meta.`environment.system`.map(_.isWow64)) ~
+      ("memory_mb" -> meta.`environment.system`.map(_.memoryMB)) ~
```
sunahsuh: I think for some of the less general-purpose fields here (env_build_arch, is_wow64) it might make sense to add them to the schema as user properties instead of adding them for every job that uses this code

jklukas: I think we'd have to add them to every event in the schema definition if we don't do it here. This seems like a simpler solution.

src/main/scala/com/mozilla/telemetry/pings/Ping.scala
jklukas: I see now the top-level fields are defined in https://amplitude.zendesk.com/hc/en-us/articles/204771828#keys-for-the-event-argument

Does seem like those should be sufficient. I'll remove these.

src/main/scala/com/mozilla/telemetry/pings/Ping.scala
sunahsuh: Well, since there aren't any other jobs running regularly right now I think that's fine but 1. we should consider adding top-level user properties that apply to all events in the schema/group as a follow-on bug or 2. seriously consider moving away from json definitions and moving toward a more flexible dsl

PR #163, opened 2018-07-30
Crash ping streaming base class and OpenTSDB streaming

Ben-Wu:


src/main/scala/com/mozilla/telemetry/streaming/CrashesToOpenTsdb.scala
```diff
@@ -0,0 +1,32 @@
...
+
+object CrashesToOpenTsdb extends CrashPingStreamingBase {
+
+  override val sparkAppName: String = this.getClass.getSimpleName
+
+  override def buildOutputString(measurementName: String, timestamp: Long,
+                                 buildId: String, tags: Map[String, String]): String = {
+    val formattedTags = tags
+      // Can't escape quotes in string interpolation: https://github.com/scala/bug/issues/6476
+      .map { case (k, v) => List(k, ": ", v).mkString("\"", "\"", "\"") }
```
akkomar: I suggest to either use triple quotes here:
```
s""""$k": "$v""""
```
or use json4s to build json from map, like in https://github.com/mozilla/telemetry-streaming/blob/master/src/main/scala/com/mozilla/telemetry/streaming/ExperimentEnrollmentsToTestTube.scala#L114-L121

src/main/scala/com/mozilla/telemetry/streaming/CrashesToOpenTsdb.scala
```diff
@@ -0,0 +1,32 @@
...
+import com.mozilla.telemetry.sinks.BatchHttpSink
+
+object CrashesToOpenTsdb extends CrashPingStreamingBase {
+
+  override val sparkAppName: String = this.getClass.getSimpleName
+
+  override def buildOutputString(measurementName: String, timestamp: Long,
+                                 buildId: String, tags: Map[String, String]): String = {
+    val formattedTags = tags
+      // Can't escape quotes in string interpolation: https://github.com/scala/bug/issues/6476
```
akkomar: I think we no longer need this comment.

src/test/scala/com/mozilla/telemetry/streaming/CrashesToOpenTsdbTest.scala
```diff
@@ -0,0 +1,51 @@
...
+    isJsonObject("[]") should be (false)
+  }
+
+  "isJsonObject" should "return true for json objects" in {
+    isJsonObject("{}") should be (true)
+    isJsonObject("{\"a\": 1}") should be (true)
+    isJsonObject("{\"a\": {\"b\": 3}}") should be (true)
+  }
+
+  "isJsonObject" should "return false for strings with unbalanced braces" in {
```
akkomar: I think we don't need to test json4s (3 `"isJsonObject"` tests are doing that).

PR #165, opened 2018-07-31
Bug 1474987 Add "session split" meta-event

jklukas:


src/test/scala/com/mozilla/telemetry/streaming/EventsToAmplitudeTest.scala
```diff
@@ -314,7 +328,7 @@ class EventsToAmplitudeTest extends FlatSpec with Matchers with BeforeAndAfterAl
     verify(expectedTotalMsgs, createMatcher(mainPingJsonMatch))
   }
 
-  "Events to Amplitude" should "ignore pings without events" taggedAs(Kafka.DockerComposeTag, DockerEventsTag, DockerMainEvents) in {
+  it should "ignore pings without events" taggedAs(Kafka.DockerComposeTag, DockerEventsTag, DockerMainEvents) in {
```
jklukas: Changing these to `it`s just to pretty up the test output.

src/main/scala/com/mozilla/telemetry/pings/MainPing.scala
```diff
@@ -107,6 +133,26 @@ case class MainPing(application: Application,
     dynamicProcessEvents.filter(_.category == "normandy")
   }
 
+
+  override def pingLevelEvents: Seq[JObject] = {
```
sunahsuh: I think it makes a lot more sense to insert an event into the main ping events list and add an entry into each individual config re: whether they'd like to use this event or not. As it stands, other jobs using this code will *always* have this event, which is problematic for a number of reasons. For instance, in the case of devtools which looks through *all* main pings, _every single main ping will now fire this event_ whether that client used devtools or not.

jklukas: Makes total sense. I'll add something to the JSON schema to enable or disable this behavior.

configs/desktop_savant_events_schemas.json
```diff
@@ -1,5 +1,6 @@
 {
   "source": "telemetry-cohorts",
+  "includePingLevelEvent": true,
```
sunahsuh: I hate to do this, but I have two suggestions:
- We can make this an event we filter on just like the other events, where category/object/method would probably be something like "meta/telemetry/sessionSplit", and we can stuff that into the ping's `events` value, and we can avoid adding another construct to the schema.
- Alternately if you want to skip that I think this would be more accurately named "includeSessionSplitEvent"

src/main/scala/com/mozilla/telemetry/pings/MainPing.scala
```diff
@@ -107,6 +133,26 @@ case class MainPing(application: Application,
...
+      ("session_id" -> sessionStart) ~
+      ("insert_id" -> (getClientId.getOrElse("None") + sessionStart.toString + "ping sent")) ~
+      ("event_type" -> "ping sent") ~
+      ("time" -> sessionStart) ~
+      ("event_properties" ->
+        JObject() ~
+          ("subsession_length" -> meta.`payload.info` \ "subsessionLength") ~
+          ("active_ticks" -> meta.`payload.simpleMeasurements` \ "activeTicks") ~
+          ("uri_count" -> getScalarValue("parent", "browser.engagement.total_uri_count")) ~
+          ("search_count" -> searchCount)
```
sunahsuh: Consider also adding the "reason" field re: session split

src/main/scala/com/mozilla/telemetry/pings/MainPing.scala
```diff
@@ -107,6 +133,26 @@ case class MainPing(application: Application,
...
 
+
+  override def pingLevelEvents: Seq[JObject] = {
+    import org.json4s.JsonDSL._
+
+    val pingSent = pingAmplitudeProperties merge
+      ("session_id" -> sessionStart) ~
+      ("insert_id" -> (getClientId.getOrElse("None") + sessionStart.toString + "ping sent")) ~
+      ("event_type" -> "ping sent") ~
+      ("time" -> sessionStart) ~
```
sunahsuh: This should probably be the ping's created time since this "event" is at the end of a session

configs/desktop_savant_events_schemas.json
jklukas: includeSessionSplitEvent indeed is a better descriptor here. I do see that the value in making this more like other events, but the top-level field seems more explicit to me. Inserting an event in the stream and then expecting schema writers to know about its coordinates if they want to include it seems more magical. I'm planning to keep it so the `eventGroups` only needs to worry about events defined in the normal way, and the session split ping is a separate path.

Either way we go, we'd end up duplicating a certain amount of code.

src/main/scala/com/mozilla/telemetry/pings/MainPing.scala
```diff
@@ -107,6 +133,28 @@ case class MainPing(application: Application,
     dynamicProcessEvents.filter(_.category == "normandy")
   }
 
+
+  override def sessionSplitEvents: Seq[JObject] = {
+    import org.json4s.JsonDSL._
+
+    val eventType = "session split"
+    val event = pingAmplitudeProperties merge
```
sunahsuh: Two more points re: top-level vs "normalish" event
- The top-level field only makes sense if main pings are included in the doctype field, which may or may not be the case for many of these jobs going forward especially as event pings go to release.  What should the behavior be if `includeSessionSplitEvent` is true but the `docType` filter doesn't include `main`? 
- There's already a bit of a divergence in logic here (sessionStart calculations here vs in `eventToAmplitudeEvent`). I can only imagine this getting worse over time. If you'd like to keep the top-level field, I think it'd be a good idea to add a comment in `eventToAmplitudeEvent` about checking over this method to make sure the changes shouldn't be reflected here. I still think that's a sign this should be reworked -- if you don't want to go with specifying an event in the schema then it would still make sense to create `Event` and `AmplitudeEvent` objects and pass them through the common formatter.

jklukas: >  What should the behavior be if includeSessionSplitEvent is true but the docType filter doesn't include main

The sessionSplitEvent is empty by default and has to be overridden specifically by different ping types. So it's well defined that behavior in this case is that no sessionSplitEvent is produced.

We'd have the same problem if we injected a special event for the session split and let it flow through. That special event needs access to fields specific to the ping type, so we'd need to define the event per ping type, likely making the default to be no event. So there would be the possibility of configuring the schema to pick up the event, but the Scala code doesn't implement any event to match, so would have no effect.

> it would still make sense to create Event and AmplitudeEvent objects and pass them through the common formatter.

I'm playing with this idea.

sunahsuh: Okay, one more thought/piece of context that might not be clear: there is no such thing as a session split in mobile. This is a concept that's specific to the main ping, hence why it feels wrong to me to add this at the top-level to a schema that includes non-desktop pings.

> So it's well defined that behavior in this case is that no sessionSplitEvent is produced.

My point is that if that happens it's likely an error by the person writing the schema. If we want to get to a point where it's not data engineers writing the schema (which, dear deity I hope we do) we should consider where we might be putting footguns ;)

docs/amplitude/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+`name` determines the name of the event as seen in Amplitude.
+
+`description` is a comment field. It doesn't get sent to Amplitude, but rather
+exists as an opportunity for you to document the purpose of the event or any
+additional caveats for future maintainters of the schema.
+
+`sessionIdOffset` is an optional field created for the `devTools` job that gives
+you a chance to identify a field to use to provide additional disambiguation
+between events.
```
jklukas: I don't really understand this field, so this description may be way off. cc @sunahsuh

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
```diff
@@ -160,7 +166,9 @@ object EventsToAmplitude extends StreamingJobBase {
         try {
           parsePing(Message.parseFrom(v.get(0).asInstanceOf[Array[Byte]]), sample, config)
         } catch {
-          case _: Throwable if !raiseOnError => Array[String]()
+          case NonFatal(ex) if !raiseOnError =>
+            log.error("Encountered an error!", ex)
+            Array[String]()
```
jklukas: I now see that we have this pattern of catching any throwable in several places, so maybe I shouldn't change this. It seems super dangerous, but maybe this is an accepted pattern in Spark?

src/main/scala/com/mozilla/telemetry/pings/MainPing.scala
```diff
@@ -107,6 +133,22 @@ case class MainPing(application: Application,
     dynamicProcessEvents.filter(_.category == "normandy")
   }
 
+  override def sessionSplitEvents: Seq[Event] = {
+    val timestamp: Int = meta.`payload.info` \ "sessionLength" match {
```
sunahsuh: Don't feel like you need to do this right now, but this would be better as a `sessionLength` getter in `Meta`

src/main/scala/com/mozilla/telemetry/pings/MainPing.scala
```diff
@@ -107,6 +133,22 @@ case class MainPing(application: Application,
...
+      case JInt(x) => x.toInt
+      case _ => 0
+    }
+    val extra: Map[String, Any] = Map(
+      "subsession_length" -> meta.`payload.info` \ "subsessionLength",
+      "active_ticks" -> meta.`payload.simpleMeasurements` \ "activeTicks",
+      "uri_count" -> getScalarValue("parent", "browser.engagement.total_uri_count"),
+      "search_count" -> searchCount,
+      "reason" -> meta.`payload.info` \ "reason"
```
sunahsuh: Same, re: getters here

src/main/scala/com/mozilla/telemetry/pings/MainPing.scala
```diff
@@ -119,10 +161,17 @@ case class MainPing(application: Application,
     // sessionStartDate is truncated to the hour anyway, so we just want to get somewhere near the correct timeframe
     case _ => ((meta.Timestamp / 1e9) - events.map(_.timestamp).max).toLong
   }
+
+  def pingSentTime: Long = meta.`payload.info` \ "sessionLength" match {
```
sunahsuh: Is this left over from before?

src/main/scala/com/mozilla/telemetry/pings/Ping.scala
```diff
@@ -297,8 +297,18 @@ trait SendsToAmplitude {
...
+  def sessionSplitEvents: Seq[Event] = Seq.empty
+
   def eventToAmplitudeEvent(eventGroup: String, e: Event, es: AmplitudeEvent): JObject = {
-    val sessionIdOffset = Try(es.sessionIdOffset.map(o => e.getField(o).toLong).getOrElse(0L)).getOrElse(0L)
+    val sessionIdOffset: Long = Try(
+      es.sessionIdOffset
+        .map(e.getField(_) match {
+          case JInt(x) => x.toLong
+          case _ => 0L
+        })).getOrElse(None).getOrElse(0L)
```
sunahsuh: If you change this to a match case statement then the `Try` is extraneous

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
sunahsuh: I think @akkomar had some opinions about this -- but the main consideration I believe was to keep streaming jobs running

jklukas:
> Are we certain that sending over numerics as strings can't be handled on the amplitude side?

I will ask Amplitude folks about that.

src/main/scala/com/mozilla/telemetry/pings/Ping.scala
```diff
@@ -401,15 +412,18 @@ case class Event(timestamp: Int,
                  method: String,
                  `object`: String,
                  value: Option[String],
-                 extra: Option[Map[String, String]]) {
+                 extra: Option[Map[String, Any]]) {
 
-  def getField(field: String): String = field match {
+  def getField(field: String): JValue = field match {
```
sunahsuh: Sorry, one more comment bc apparently I missed this change -- what's the reasoning re: the decision to make this return a JValue?

docs/amplitude/README.md
```diff
@@ -0,0 +1,137 @@
+# Amplitude Event Configuration
```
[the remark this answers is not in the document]
jklukas: Fine idea. Will do.

docs/amplitude/README.md
jklukas: I like it. Will incorporate.

src/main/scala/com/mozilla/telemetry/pings/MainPing.scala
jklukas: Probably. Will check.

src/main/scala/com/mozilla/telemetry/pings/Ping.scala
jklukas: This is related to extra changing to `Map[String, Any]`. We were previously relying on implicit conversions from String to JValue when using this function. That doesn't work generically if the return type here is `Any`, so changing the output type to `JValue` lets the implicit conversion happen within this function for most of the strings, and we explicitly convert where necessary.

PR #166, opened 2018-08-06
Bug 1469870 Move from TravisCI to CircleCI 2.0

jklukas:
This moves tests out of TravisCI to CircleCI 2.0.

As part of this change, we split work into multiple jobs: linting, main tests, tests with Kafka, assembly, and deploy. [The entire workflow runs in ~6 minutes](https://circleci.com/workflow-run/791e7aa7-0611-4f7f-91f0-0a50d02affa3) compared to ~10 minutes on TravisCI.

For the tests that require Kafka, I considered two options. One was to use the `machine` executor type, install docker-compose, and call the existing `sbt dockerComposeTest`. The other was to avoid docker-compose and instead use CircleCI's native support for "secondary" containers running databases or other dependencies.

I chose to use secondary containers rather than docker-compose because the configuration ended up being simpler and more similar to the other jobs in the workflow; it also runs more quickly. The docker-compose configuration is now only for local testing. That introduces some duplication of logic and could lead to skew between what runs in CI and what runs locally, but I don't anticipate we'll often need to change the docker-compose configuration.


build.sbt
```diff
@@ -74,9 +74,6 @@ test in assembly := {}
 // Add configs to resources
 unmanagedResourceDirectories in Compile += baseDirectory.value / "configs"
 
-// Default SBT settings suck
-javaOptions ++= Seq("-Xms512M", "-Xmx2048M", "-XX:MaxPermSize=2048M", "-XX:+CMSClassUnloadingEnabled")
```
jklukas: Factored these out to .jvmopts to match what t-b-v does.

jklukas:
> One thing I think we're missing is clear instruction on how to locally run the same set of tests/checks as on the CI (previously there was sbt ci).

That's a reasonable point. I restored the `ci` task and added a comment and updated the doc to try to make it more clear that `sbt ci` is not _literally_ what gets run by CI, but provides a local method of running the same set of tests.

jklukas:
Updated requirements for the `master` branch to reference CircleCI tests rather than TravisCI, so this PR now shows as green for CI.

PR #167, opened 2018-08-07
Add experiment enrollments to datadog job

sunahsuh:
WIP while I make sure this new version works in databricks

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/monitoring/DogStatsDMetric.scala
+++ b/src/main/scala/com/mozilla/telemetry/monitoring/DogStatsDMetric.scala
@@ -1,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package com.mozilla.telemetry.monitoring
+
+
+/* Conforms to the DogStatsD datagram format as documented here:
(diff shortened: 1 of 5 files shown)


sunahsuh:
Restarted the job on databricks and it's chugging along -- also fixed the Sink test because apparently it was not doing anything and exiting before the Future completed

src/main/scala/com/mozilla/telemetry/monitoring/DogStatsDMetric.scala
```diff
@@ -0,0 +1,41 @@
...
+    Array(Some(metric), Some(metricType), sampleRateString(sampleRate), tags)
+      .flatten
+      .mkString("|")
+  }
+
+  private def normalize(input: String): String = {
+    input.replaceAll("[:|@]", "_")
+  }
+
+  private def sampleRateString(sampleRate: Option[Double]): Option[String] = sampleRate.map("@" + _.toString)
```
akkomar: nit: I'd add a comment pointing to Datadog documentation to explain magic `@` here and others in line 20 (https://github.com/mozilla/telemetry-streaming/pull/167/files#diff-ebf9e4f9305826d7c4e93f152216a9e0R20)

sunahsuh: Makes sense -- I added a link to the docs at the top of the trait and added a comment on the `normalize` method about the reserved characters

PR #171, opened 2018-08-13
Bug 1482924 Use active_ticks scalar rather than simpleMeasurement

jklukas:
The simpleMeasurement version of activeTicks is sometimes inflated.

We will need to backfill the Savant project in Amplitude using this updated logic.

### DIFF
--- a/src/test/scala/com/mozilla/telemetry/TestUtils.scala
+++ b/src/test/scala/com/mozilla/telemetry/TestUtils.scala
@@ -205,7 +205,7 @@ object TestUtils {
       "payload.simpleMeasurements" ->
         """
           |{
-          |  "activeTicks": 2000,
+          |  "activeTicks": 275,
           |  "firstPaint": 1200
           |}
         """.stripMargin,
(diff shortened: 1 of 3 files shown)


src/test/scala/com/mozilla/telemetry/TestUtils.scala
```diff
@@ -205,7 +205,7 @@ object TestUtils {
       "payload.simpleMeasurements" ->
         """
           |{
-          |  "activeTicks": 2000,
+          |  "activeTicks": 275,
```
jklukas: Putting in a value slightly above 271 here to mimic the activeTicks overcounting bug. This number should never show up in the tests, because the scalar version should get picked up instead.

src/test/scala/com/mozilla/telemetry/streaming/EventsToAmplitudeTest.scala
```diff
@@ -96,7 +96,7 @@ class EventsToAmplitudeTest extends FlatSpec with Matchers with BeforeAndAfterAl
        |  "event_type": "Meta - session split",
        |  "event_properties": {
        |    "subsession_length": "3600",
-       |    "active_ticks": "2000",
+       |    "active_ticks": "271",
```
jklukas: The change the expected value here to match what we were injecting as the scalar value for active_ticks.

PR #173, opened 2018-08-16
Bug 1482924 Add event name to insert_id

jklukas:
The "Meta - session split v2" events we backfilled were
rejected for recent days due to having identical insert_id
to the v1 events.

We change the logic so that future event name changes will
not get deduplicated, and we bump the session split event name again.

This time, I'll backfill and verify _before_ merging the PR in case
we encounter another hiccup.


jklukas:
Ran the backfill and event counts for the new event look good all the way up to last night.

PR #178, opened 2018-08-22
Bug 1471368: Update meta-schema file and config files

sunahsuh:
As it turns out, we were never checking the config validation output in our validation code -- fixing this uncovered a bunch of issues with the meta-schema drifting over time from what was in the code, and sundry small issues in the config files that shouldn't have code impact.

I highly suggest using the `?w=1` trick at the end of the PR's url to hide whitespace changes

### DIFF
--- a/configs/focus_android_events_schemas.json
+++ b/configs/focus_android_events_schemas.json
@@ -1,2209 +1,2070 @@
 {
     "source": "telemetry",
     "eventGroupName": "Focus",
-    "events": [
-        [
-            {
-                "amplitudeProperties": {}, 
-                "description": "", 
-                "name": "Start session", 
-                "schema": {
-                    "$schema": "http://json-schema.org/schema#", 
-                    "properties": {
-                        "category": {
-                            "enum": [
-                                "action"
-                            ], 
-                            "type": "string"
-                        }, 
-                        "method": {
-                            "enum": [
-                                "foreground"
-                            ], 
-                            "type": "string"
-                        }, 
-                        "object": {
-                            "enum": [
-                                "app"
-                            ], 
-                            "type": "string"
-                        }, 
-                        "timestamp": {
-                            "minimum": 0, 
-                            "type": "number"
-                        }
-                    }, 
-                    "required": [
-                        "category", 
-                        "timestamp", 
-                        "object", 
-                        "method"
-                    ], 
-                    "type": "object"
+    "eventGroups": [
+        {
+            "eventGroupName": "m_foc",
+            "events": [
+                {
+                    "amplitudeProperties": {},
+                    "description": "",
+                    "name": "Start session",
+                    "schema": {
+                        "$schema": "http://json-schema.org/schema#",
+                        "properties": {
+                            "category": {
+                                "enum": [
+                                    "action"
+                                ],
+                                "type": "string"
+                            },
+                            "method": {
+                                "enum": [
+                                    "foreground"
+                                ],
+                                "type": "string"
+                            },
+                            "object": {
+                                "enum": [
+                                    "app"
+                                ],
+                                "type": "string"
+                            },
+                            "timestamp": {
+                                "minimum": 0,
+                                "type": "number"
+                            }
+                        },
+                        "required": [
+                            "category",
+                            "timestamp",
+                            "object",
+                            "method"
+                        ],
+                        "type": "object"
+                    }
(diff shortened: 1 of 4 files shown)


configs/focus_android_events_schemas.json
```diff
@@ -1,2209 +1,2070 @@
 {
     "source": "telemetry",
     "eventGroupName": "Focus",
```
jklukas: I don't think this top-level eventGroupName should be here anymore.

jklukas:
I just realized I wasn't the assigned reviewer here; I saw this through bugzilla updates. But you got my opinion anyway.

configs/focus_android_events_schemas.json
```diff
@@ -1,2209 +1,2070 @@
...
-                    "required": [
-                        "category", 
-                        "timestamp", 
-                        "object", 
-                        "method"
-                    ], 
-                    "type": "object"
+    "eventGroups": [
+        {
+            "eventGroupName": "m_foc",
```
fbertsch: This will change the event names, no? Shouldn't this be `Focus`?

src/main/resources/schemas/schemaFileSchema.json
```diff
@@ -72,21 +76,33 @@
...
         }
       },
       "additionalProperties": false
     },
     "eventGroups": {
       "type": "array",
+      "minItems": 1,
       "items": {
         "title": "Event group",
         "type": "object",
```
fbertsch: Thoughts on adding `"additionalProperties": false` to this object to prevent other kinds of drift?

src/main/resources/schemas/schemaFileSchema.json
```diff
@@ -140,6 +160,10 @@
                       {
                         "type": "string",
                         "pattern": "^extra"
+                      },
+                      {
+                        "type": "string",
+                        "pattern": "^literal"
```
fbertsch: We could just merge this with the above, `^(extra|literal)`. Up to you, no change required here.

src/main/resources/schemas/schemaFileSchema.json
```diff
@@ -118,6 +134,10 @@
                       {
                         "type": "string",
                         "pattern": "^extra"
+                      },
```
fbertsch: Really we should merge the `userProperties` and `amplitudeProperties` into a single reference (unless there are differences I missed).

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
```diff
@@ -187,7 +187,10 @@ object EventsToAmplitude extends StreamingJobBase {
     // validate config json
     val factory = JsonSchemaFactory.byDefault
     val schema = factory.getJsonSchema(asJsonNode(getMetaSchema))
-    schema.validate(asJsonNode(json))
+    val report = schema.validate(asJsonNode(json))
```
fbertsch: Did this stop throwing an exception on non-validation? I could have _sworn_ I made it explicitly do that...

src/main/resources/schemas/schemaFileSchema.json
jklukas: amplitudeProperties lead to entries in the event_properties dictionary in the event payload while userProperties lead to entries in the user_properties dictionary. There are differences in how those two types of properties are presented in the Amplitude UI for querying.

fbertsch: Right, what I'm referring to is the duplicated JSON schema code here. We can create a reference at the top of the schema file which can then be used for both of these properties, if they match.

sunahsuh: Will do, but in the process of doing this I noticed -- the definitions that are already there seem unused except `stringArray` and `nonNegativeInteger` (which is used in another, unused definition)

src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala
sunahsuh: `validate` *does* throws its own exception but really only when the json isn't parseable (or anything else it deems a "processing error"). Otherwise it'll spit out a ProcessingReport, which is an iterable of messages, which in this case is essentially the list of validation errors, so that return value has to be examined.

PR #179, opened 2018-08-23
Bug 1485583 - Add accumulator-based metrics source

akkomar:
This adds simple Spark metrics source based on accumulators.

Because it's leveraging internal API, source is added in `org.apache.spark`.

### DIFF
--- a/src/main/scala/org/apache/spark/metrics/source/custom/AccumulatorMetricsSource.scala
+++ b/src/main/scala/org/apache/spark/metrics/source/custom/AccumulatorMetricsSource.scala
@@ -1,0 +1,94 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package org.apache.spark.metrics.source.custom
+// This class is defined under `org.apache.spark` package, because it extends
+// package-private `org.apache.spark.metrics.source.Source`
+
+import java.util.concurrent.{Executors, TimeUnit}
+import java.{lang => jl}
+
+import com.codahale.metrics.MetricRegistry
+import org.apache.commons.lang3.concurrent.BasicThreadFactory
+import org.apache.log4j.LogManager
+import org.apache.spark.SparkEnv
+import org.apache.spark.metrics.source.Source
+import org.apache.spark.util.{AccumulatorV2, ShutdownHookManager}
+
+import scala.collection.mutable
+
+/**
+  * Accumulator-based metric source.
+  *
+  * This source uses Spark accumulators for collecting metrics. After registering an accumulator, it will be
+  * periodically polled for its value which will be reported to Spark metrics system using [[com.codahale.metrics.Meter]].
+  * After accumulators are registered it is necessary to start the source with [[AccumulatorMetricsSource#start()]],
+  * which will start internal polling thread and register shutdown hook for flushing metrics.
+  *
+  * Usage:
+  * {{{
+  * val metricsSource = new AccumulatorMetricsSource("prefix")
+  * val metric = spark.sparkContext.longAccumulator("metric-name")
+  * metricsSource.registerAccumulator(metric)
+  * metricsSource.start()
+  * }}}
+  *
+  * @param prefix for metrics collected by this source
+  */
+class AccumulatorMetricsSource(prefix: String) extends Source {
+  override val sourceName: String = prefix
+  override val metricRegistry: MetricRegistry = new MetricRegistry
+
+  private val log = LogManager.getLogger(this.getClass.getName)
+
+  private val accumulators = mutable.Map[String, AccumulatorV2[jl.Long, jl.Long]]()
+
+  private val executor = Executors.newSingleThreadScheduledExecutor(
+    new BasicThreadFactory.Builder()
+      .namingPattern("accumulator-metrics-source-thread-%d")
+      .daemon(true)
+      .priority(Thread.NORM_PRIORITY)
+      .build())
+
+  def registerAccumulator(acc: AccumulatorV2[jl.Long, jl.Long]): Unit = {
+    val name = acc.name.getOrElse("UNNAMED_ACCUMULATOR")
+    require(!accumulators.keySet.contains(name), "Accumulator already registered")
+    accumulators.put(name, acc)
+    metricRegistry.meter(name) // register metric beforehand to make sure no errors are thrown later during collection
+  }
+
+  def start(pollingPeriod: Long = 10, unit: TimeUnit = TimeUnit.SECONDS): Unit = {
+    executor.scheduleAtFixedRate(new Runnable() {
+      override def run(): Unit = {
+        try {
+          report()
+        } catch {
+          case ex: Exception =>
+            log.error("Exception thrown from AccumulatorMetricsSource#report. Exception was suppressed.", ex)
+        }
+      }
+    }, pollingPeriod, pollingPeriod, unit)
+
+    SparkEnv.get.metricsSystem.registerSource(this)
+    ShutdownHookManager.addShutdownHook(() => this.stop())
+  }
+
+  private[custom] def stop(): Unit = {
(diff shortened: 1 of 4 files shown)


src/main/scala/org/apache/spark/metrics/source/custom/AccumulatorMetricsSource.scala
```diff
@@ -0,0 +1,54 @@
...
+  override def sourceName: String = prefix
+  override def metricRegistry: MetricRegistry = registry
+
+  private val registry = new MetricRegistry
+  private val gauges = mutable.Map[String, Long]()
+  private val meters = mutable.Map[String, Meter]()
+
+  def registerAccumulator(acc: AccumulatorV2[jl.Long, jl.Long], name: String) {
+    gauges.put(name, 0L)
+    meters.put(name, metricRegistry.meter(name + "-rate"))
```
jklukas: Is it necessary to maintain the mutable maps here? I think we could instantiate the gauge and the meter directly and reference them in the gauge's `getValue` method without having to look them up in a map.

src/main/scala/org/apache/spark/metrics/source/custom/AccumulatorMetricsSource.scala
```diff
@@ -0,0 +1,54 @@
...
+
+    metricRegistry.register(name,
+      new Gauge[Long] {
+        override def getValue: Long = {
+          val value = acc.value
+          meters(name).mark(getCountOverflowSafe(value, gauges(name)))
```
jklukas: This feels like it's abusing the intent of `getValue` somewhat. Ideally, this function should be simply accessing a precomputed value, but we're including a side effect of updating the meter.

Is it possible for us to provide a mark method that jobs would call directly? Or is the idea here that we need to keep the metrics themselves on the master node so that they don't get serialized and copied on workers?

I certainly don't want us to spin up a thread executor to handle periodically marking the meter, so maybe this is the best we can do.

src/main/scala/org/apache/spark/metrics/source/custom/AccumulatorMetricsSource.scala
```diff
@@ -0,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package org.apache.spark.metrics.source.custom
```
jklukas: I see we're overriding a class that's private to spark. How worried should we be about that?

We should provide some clear commentary here about why we're putting a class under the spark package name and the potential danger of doing so.

src/main/scala/org/apache/spark/metrics/source/custom/AccumulatorMetricsSource.scala
```diff
@@ -0,0 +1,54 @@
...
+  * metricsSource.registerAccumulator(metric, metric.name.getOrElse("UNNAMED_ACCUMULATOR"))
+  * SparkEnv.get.metricsSystem.registerSource(metricsSource)
+  * }</pre>
+  * @param prefix for metrics collected by this source
+  */
+class AccumulatorMetricsSource(prefix: String) extends Source {
+  override def sourceName: String = prefix
+  override def metricRegistry: MetricRegistry = registry
+
+  private val registry = new MetricRegistry
```
jklukas: Any reason to make this separate rather than setting metricRegistry directly?

```
override def metricRegistry: MetricRegistry = new MetricRegistry
```

src/main/scala/org/apache/spark/metrics/source/custom/AccumulatorMetricsSource.scala
```diff
@@ -0,0 +1,54 @@
...
+  private val gauges = mutable.Map[String, Long]()
+  private val meters = mutable.Map[String, Meter]()
+
+  def registerAccumulator(acc: AccumulatorV2[jl.Long, jl.Long], name: String) {
+    gauges.put(name, 0L)
```
jklukas: I don't think we need a gauge at all. A Dropwizard Meter emits five separate metrics:

- m1, m5, and m15 rates
- mean rate
- [count](https://github.com/dropwizard/metrics/blob/3.2-maintenance/metrics-core/src/main/java/com/codahale/metrics/Meter.java#L79-L82)

I believe the gauge here is redundant with the meter's `count`.

jklukas:
Is accumulating necessary here? Would it be better to have each task report metrics individually? Then we can aggregate across nodes in the DataDog dashboard or whatever other visualization tool?

Are accumulators updated frequently, or do they only report back to the master node when the task finishes?

Are there factors that would prevent us from directly instantiating a Meter inside a task and letting that node report its own metrics? Is this PR operating under the assumption that we want to minimize the number of reported metrics in order to reduce cost?

jklukas:
> Are there factors that would prevent us from directly instantiating a Meter inside a task and letting that node report its own metrics?

I just realized that without wrapping this in a Source and registering it, the metrics would never get reported anywhere. I suppose we would have to maintain a separate thread for sending the metrics to statsd, or maintain a separate process to query them. Neither of those sound good. So integrating with Spark's metrics is seeming to be a good bargain now.

src/main/scala/org/apache/spark/metrics/source/custom/AccumulatorMetricsSource.scala
akkomar: That would return new instance at each call. Nevertheless, that's a good point - I'll switch these `def`s to `val`s.

src/main/scala/org/apache/spark/metrics/source/custom/AccumulatorMetricsSource.scala
akkomar: :+1: adding a comment.
It's not dangerous, we're extending `Source` trait here - if its API changes in the future, compiler will warn us.

akkomar:
@jklukas thanks for all the questions, they got me thinking that this approach might be too "hacky". Please take another look - I refactored this a bit. Although the code is a little longer, I think now it's easier to understand and doesn't rely on non-obvious mechanisms from underlying libraries.

Previously, gauge was used solely for the purpose of triggering updates (as this is the only metric type that can "pull" its value) and as you noticed, it was redundant. I got rid of it and used scheduled task for polling accumulators (another benefit of that is ability to control polling period independently from Spark's metric sink one).

> Is accumulating necessary here? Would it be better to have each task report metrics individually? Then we can aggregate across nodes in the DataDog dashboard or whatever other visualization tool?

We could, but then we would have to maintain MetricRegistry on each executor.

> Are accumulators updated frequently, or do they only report back to the master node when the task finishes?

They're updated after task is finished.

src/main/scala/org/apache/spark/metrics/source/custom/AccumulatorMetricsSource.scala
```diff
@@ -0,0 +1,94 @@
...
+      .namingPattern("accumulator-metrics-source-thread-%d")
+      .daemon(true)
+      .priority(Thread.NORM_PRIORITY)
+      .build())
+
+  def registerAccumulator(acc: AccumulatorV2[jl.Long, jl.Long]) {
+    val name = acc.name.getOrElse("UNNAMED_ACCUMULATOR")
+    require(!accumulators.keySet.contains(name), "Accumulator already registered")
+    accumulators.put(name, acc)
+    metricRegistry.meter(name)
```
jklukas: This method should define return type `: Unit`.

Also,  is this call to `metricRegistry.meter(name)` necessary? Do we want to make sure the meter is created at this point? Since we don't use the return value, a comment about intention would be helpful.

akkomar: > This method should define return type : Unit.

:+1: 

> Also, is this call to metricRegistry.meter(name) necessary? Do we want to make sure the meter is created at this point? Since we don't use the return value, a comment about intention would be helpful.

Adding a comment here, it's better to keep it in order to catch registration errors as early as possible.

PR #182, opened 2018-09-20
Bug 1465974 Add metrics reporting to HttpSink

jklukas:
Factors some test fixtures out of HttpSinkTest and AccumulatorMetricsSourceTest in order to reuse them in the new test for HttpSink.Metrics

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/sinks/HttpSink.scala
+++ b/src/main/scala/com/mozilla/telemetry/sinks/HttpSink.scala
@@ -3,9 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 package com.mozilla.telemetry.sinks
 
-import org.apache.spark.SparkContext
 import org.apache.spark.metrics.source.custom.AccumulatorMetricsSource
-import org.apache.spark.sql.ForeachWriter
+import org.apache.spark.sql.{ForeachWriter, SparkSession}
 import org.apache.spark.util.LongAccumulator
 import scalaj.http.{Http, HttpRequest}
 
@@ -32,15 +31,15 @@ object HttpSink {
     metrics: Option[Metrics] = None
   ) {
     /** Return a copy of this config with HttpSink metrics enabled. */
-    def withMetrics(implicit sc: SparkContext): Config = {
+    def withMetrics(implicit spark: SparkSession): Config = {
       val source = new AccumulatorMetricsSource("HttpSink")
-      val config = this.withMetrics(source)
+      val config = this.withMetrics(source)(spark)
       source.start()
       config
     }
 
     /** Return a copy of this config with HttpSink metrics enabled, registering accumulators with the given source. */
-    def withMetrics(source: AccumulatorMetricsSource)(implicit sc: SparkContext): Config = {
+    def withMetrics(source: AccumulatorMetricsSource)(implicit spark: SparkSession): Config = {
       require(this.metrics.isEmpty, "Metrics are already enabled for this HttpSink.Config")
       val metrics = Metrics.registeredTo(source)
       this.copy(metrics = Option(metrics))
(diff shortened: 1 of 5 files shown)


src/test/scala/org/apache/spark/metrics/source/custom/AccumulatorMetricsSourceTest.scala
```diff
@@ -14,13 +14,43 @@ import org.scalatest._
...
-  private val MetricName = "test-metric"
+  override protected def beforeEach(): Unit = {
+    cleanupTestDirectories()
+    Files.createDirectories(Paths.get(MetricsOutputDir))
+    super.beforeEach()
+  }
+
+  override protected def afterEach(): Unit = {
+    try super.afterEach()
+    finally cleanupTestDirectories()
```
akkomar: Good catch with `before-` and `afterEach` implementations.

src/test/scala/com/mozilla/telemetry/sinks/HttpSinkMetricsTest.scala
```diff
@@ -0,0 +1,92 @@
...
+
+    super.beforeEach()
+  }
+
+  override protected def afterEach(): Unit = {
+    try super.afterEach()
+    finally spark.stop()
+  }
+
+  "HttpSink.Metrics" should "output metrics" in {
```
akkomar: I'm wondering if this test doesn't go too far with mixing testing of metrics source (which is already covered elsewhere - there's some copy-pasta from base accumulator metric test) and http sink behavior.

The new part it's testing (that hasn't been tested already) is accumulator usage in HttpSink (whether they're incremented correctly). Maybe cleaner way for testing this would be to only check accumulator values in the end?

src/main/scala/com/mozilla/telemetry/sinks/HttpSink.scala
```diff
@@ -12,16 +15,77 @@ import scala.util.{Failure, Success, Try}
...
     connectionTimeoutMillis: Int = 2000,
     readTimeoutMillis: Int = 5000,
     successCodes: Set[Int] = Set(OK),
-    retryCodes: Set[Int] = RetryCodes
-  )
+    retryCodes: Set[Int] = RetryCodes,
+    metrics: Option[Metrics] = None
+  ) {
+    /** Return a copy of this config with HttpSink metrics enabled. */
+    def withMetrics(implicit sc: SparkContext): Config = {
```
akkomar: I'd opt for exposing `SparkSession` in interfaces - we'll end up with a little less code when calling them.

src/main/scala/com/mozilla/telemetry/sinks/HttpSink.scala
```diff
@@ -12,16 +15,77 @@ import scala.util.{Failure, Success, Try}
...
+      val source = new AccumulatorMetricsSource("HttpSink")
+      val config = this.withMetrics(source)
+      source.start()
+      config
+    }
+
+    /** Return a copy of this config with HttpSink metrics enabled, registering accumulators with the given source. */
+    def withMetrics(source: AccumulatorMetricsSource)(implicit sc: SparkContext): Config = {
```
akkomar: `SparkSession`, as above?

src/main/scala/com/mozilla/telemetry/sinks/HttpSink.scala
jklukas: Seems reasonable. I'll use implicit SparkSession consistently in this chain.

src/test/scala/com/mozilla/telemetry/sinks/HttpSinkMetricsTest.scala
jklukas: Good point. I agree checking just accumulator values should be sufficient here.

PR #183, opened 2018-09-30
Add Android Rocket eventsToAmplitude config

cnevinc:


src/main/scala/com/mozilla/telemetry/pings/Ping.scala
```diff
@@ -392,6 +392,7 @@ object SendsToAmplitude {
   def apply(message: Message): SendsToAmplitude = {
     message.fieldsAsMap.get("docType") match {
       case Some("focus-event") => FocusEventPing(message)
+      case Some("rocket-event") => RocketEventPing(message)
```
fbertsch: This won't work, unfortunately. The messages being sent from Rocket do have a docType of `focus-event`. If you match on both the `docType` and the `appName`, that could do the trick.

fbertsch:
@cnevinc also the tests are failing right now: `[error] /telemetry-streaming/src/main/scala/com/mozilla/telemetry/pings/RocketEvent.scala:45:12: FocusSettings is already defined as case class FocusSettings`.

Let me know if you're having trouble running the tests!

cnevinc:
@fbertsch  I've updated the schema after reviewing the data from Oct 20~23. Could you please help dump the data from 11/8~11/12 for testing? Thank you!

PR #189, opened 2018-11-28
Bug 1505583: add separate devtools release channel event schema for independent sampling

sunahsuh:
Adds a version of the devtools event schema for the release channel -- duplicating the rest of the definition is not ideal, but the way refs work in jsonschema, breaking out the event groups into a separate file and importing it is non-trivial, unfortunately (refs work fine for importing whole object definitions but aren't designed for anything else, like a list.) Something to consider for e2a v2 on GCP.

There will be a corresponding PR in telemetry-airflow that will sample the release channel at 50%


sunahsuh:
> just to ensure I'm understanding: the release schema is exactly the same as the prerelease, but with a different channel. This is needed because sampling is a configuration option, and we don't want to sample prerelease.

Yes, that's correct

> Did you scope out the work to instead add separate sampling per-group, so it's one schema? That might be a bit much though and I'm happy with instead filing a bug about the issue instead so it's on our radar.

I think that'd be kind of a kludge, and my preference would be to take this into consideration in a larger refactor of how we ship events to amplitude once we're in gcp-land.

> We could also use the cmake templating we use in mozilla-pipeline-schemas.... hmmm.... ;)

I looked at that to see how we're doing it there.. but my thoughts here are that 1. we should not be using jsonschema if we're not validating json with the schema 2. we'd have to add cmake to the build chain in our scala repo? 😭

PR #224, opened 2019-02-05
Add iOS event schema and MobileEventPing

akkomar:
This adds iOS Fennec/Amplitude integration

### DIFF
--- a/configs/fennec_ios_events_schemas.json
+++ b/configs/fennec_ios_events_schemas.json
@@ -1387,7 +1387,7 @@
                 {
                     "amplitudeProperties": {}, 
                     "description": "", 
-                    "name": "Tapping Website Data in Data Management Menu", 
+                    "name": "Tap Website Data in Data Management Menu", 
                     "schema": {
                         "$schema": "http://json-schema.org/schema#", 
                         "properties": {
@@ -1426,7 +1426,7 @@
                 {
                     "amplitudeProperties": {}, 
                     "description": "", 
-                    "name": "Tapping on the Search Bar in Website Data", 
+                    "name": "Tap on the Search Bar in Website Data", 
                     "schema": {
                         "$schema": "http://json-schema.org/schema#", 
                         "properties": {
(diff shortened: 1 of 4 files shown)


src/main/scala/com/mozilla/telemetry/pings/MobileEvent.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  def getCreated: Option[Long] = Some(created)
+
+  override def pingAmplitudeProperties: JObject = {
+    ("device_id" -> getClientId) ~
+    ("device_model" -> device) ~
+    ("arch" -> arch) ~
+    ("locale" -> locale) ~
+    ("user_properties" ->
+      ("pref_defaultSearchEngine" -> settings.defaultSearchEngine) ~
```
fbertsch: Just want to point out that if it's easier for the PMs, you could make these pretty-printed names instead (e.g. `Default-Search-Engine` or similar). I'm not sure if spaces are allowed.

configs/fennec_ios_events_schemas.json
```diff
@@ -0,0 +1,1800 @@
...
+                    "amplitudeProperties": {}, 
+                    "description": "", 
+                    "name": "Tapping Bookmarks button in New Tab Settings", 
+                    "schema": {
+                        "$schema": "http://json-schema.org/schema#", 
+                        "properties": {
+                            "category": {
+                                "enum": [
+                                    "action"
+                                ],
```
fbertsch: Is `OTHER` needed?

fbertsch: nit: `Tapped`? Here and elsewhere

akkomar: :+1:, it's not.

src/main/scala/com/mozilla/telemetry/pings/MobileEvent.scala
akkomar: Hey @justindarc - would you like some nice, normalized property names here [1]?
I copied these 1:1 from your telemetry data as there seem to be couple of naming conventions in use, but maybe pretty-printing them as Frank suggests would make your life easier?

[1] https://github.com/mozilla/telemetry-streaming/blob/ios-amplitude/src/main/scala/com/mozilla/telemetry/pings/MobileEvent.scala#L42-L59

justindarc: Yes, if you could map them to pretty-printed labels, that would probably be easier for the PMs. Do you need me to provide you the labels or can you infer them from what's there?

PR #228, opened 2019-03-19
WIP: Use named column 'value' when extracting Heka message from Kafka

sunahsuh:
Submitting a PR for easy artifact deployment to s3


src/main/scala/com/mozilla/telemetry/streaming/ExperimentsEnrollmentsToDatadog.scala
```diff
@@ -60,7 +60,7 @@ object ExperimentsEnrollmentsToDatadog extends StreamingJobBase {
 
     messages.flatMap(v => {
       try {
-        val m = Message.parseFrom(v.get(0).asInstanceOf[Array[Byte]])
+        val m = Message.parseFrom(v.getAs("value").asInstanceOf[Array[Byte]])
```
akkomar:
```suggestion
        val m = Message.parseFrom(v.getAs[Array[Byte]]("value"))
```

PR #229, opened 2019-03-21
Patch for federated learning v2 experiment

crankycoder:
This patch changes a couple things:

The filtering has been updated to match the branch definitions in https://github.com/mozilla/federated-learning-v2-study-addon/blob/master/src/feature.js#L6

* Any dogfooding or training branch will be included for inbound pings.
* Training pings where models are never downloaded from S3 will have the modelVersion clobbered with Int.MaxValue 
* Test cases for each of the branches has been added with a modelVersion that reflects whether or not the client will submit a -1 value based on the `validation` flag for the branch

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
@@ -64,10 +64,15 @@ object FederatedLearningSearchOptimizer extends StreamingJobBase {
         val docType = fields.getOrElse("docType", "").asInstanceOf[String]
         if ("frecency-update" == docType) {
           val ping = FrecencyUpdatePing(m)
-          if ((ping.payload.study_variation contains "training") && (ping.payload.bookmark_and_history_num_suggestions_displayed > -1)) {
+          if (((ping.payload.study_variation contains "training") || 
+               (ping.payload.study_variation.startsWith("dogfooding")))
+            && (ping.payload.bookmark_and_history_num_suggestions_displayed > -1)) {
+
+            val model_version  = if (ping.payload.model_version == -1) Int.MaxValue else ping.payload.model_version
+
             Option(FrecencyUpdate(
               new Timestamp(clock.millis()),
-              ping.payload.model_version,
+              model_version,
               ping.payload.loss,
               ping.payload.update,
               ping.meta.clientId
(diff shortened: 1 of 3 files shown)


crankycoder:
r? @motin

Please check that the testcases match up to what we are expecting in the branch definitions for the addon.

src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
```diff
@@ -64,10 +64,15 @@ object FederatedLearningSearchOptimizer extends StreamingJobBase {
         val docType = fields.getOrElse("docType", "").asInstanceOf[String]
         if ("frecency-update" == docType) {
           val ping = FrecencyUpdatePing(m)
-          if ((ping.payload.study_variation contains "training") && (ping.payload.bookmark_and_history_num_suggestions_displayed > -1)) {
+          if (((ping.payload.study_variation contains "training") ||
+               (ping.payload.study_variation.startsWith("dogfooding")))
+            && (ping.payload.bookmark_and_history_num_suggestions_displayed > -1)) {
+
+            val model_version  = if (ping.payload.model_version == -1) Int.MaxValue else ping.payload.model_version
```
sunahsuh: This still won't cause the datapoints to be included in the model update -- [the `reduceOption` clause in `addBatch`](https://github.com/mozilla/telemetry-streaming/blob/master/src/main/scala/com/mozilla/telemetry/learning/federated/FederatedLearningSearchOptimizerS3Sink.scala#L45) means it's taking the averages from a single minimum model number that passed the previous filter

src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
```diff
@@ -64,10 +64,15 @@ object FederatedLearningSearchOptimizer extends StreamingJobBase {
         val docType = fields.getOrElse("docType", "").asInstanceOf[String]
         if ("frecency-update" == docType) {
           val ping = FrecencyUpdatePing(m)
-          if ((ping.payload.study_variation contains "training") && (ping.payload.bookmark_and_history_num_suggestions_displayed > -1)) {
+          if (((ping.payload.study_variation contains "training") ||
+               (ping.payload.study_variation.startsWith("dogfooding")))
+            && (ping.payload.bookmark_and_history_num_suggestions_displayed > -1)) {
```
motin: To match v1, the constraint should be `bookmark_and_history_rank_selected > -1`. It is currently `bookmark_and_history_num_suggestions_displayed > -1`

PR #231, opened 2019-03-25
Patch to add a modelBranch argument to the Federated Learning v2 study

crankycoder:
The Federated Learning v2 experiment needs to have an extra parameter
passed in to distinguish between the 4 different models that need to be
trained.

The filter to select relevant pings has been updatd.

Extra test case added demonstrating filtering the aggregation step by
using the variation name to drop ignorable pings.

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
@@ -74,7 +74,13 @@ object FederatedLearningSearchOptimizer extends StreamingJobBase {
         val docType = fields.getOrElse("docType", "").asInstanceOf[String]
         if ("frecency-update" == docType) {
           val ping = FrecencyUpdatePing(m)
-          if ((ping.payload.study_variation startsWith variationTarget) && (ping.payload.bookmark_and_history_num_suggestions_displayed > -1)) {
+          if (
+              (
+               (ping.payload.study_variation startsWith variationTarget) &&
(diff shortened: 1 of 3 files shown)


crankycoder:
@sunahsuh  As discussed in our chat - I've updated the filter to match the configuration in https://github.com/mozilla/federated-learning-v2-study-addon/blob/master/src/feature.js#L6

In a nutshell - just check the variationName so that : 
 * the prefix is 'model<N>' 
 * variationName does not contain "not-submitting"

src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
```diff
@@ -40,31 +40,47 @@ object FederatedLearningSearchOptimizer extends StreamingJobBase {
...
+  def aggregate(pings: DataFrame, modelBranch: Int, clock: Clock, windowOffsetMin: Int, raiseOnError: Boolean = false): Dataset[FrecencyUpdateAggregate] = {
     import pings.sparkSession.implicits._
 
+
+    val variationTarget: String = modelBranch match {
+      case 1 => "model1"
+      case 2 => "model2"
+      case 3 => "model3"
+      case 4 => "model4"
+    }
```
sunahsuh: What's the value in using this int to String mapping? This mapping is done only once -- seems simpler to just pass in the actual string to the job

crankycoder: There isn't one. :)

I've simplified this in fa3787d to use the full model string name instead of doing the conversion from int.

PR #233, opened 2019-03-31
added option for failOnDataLoss to be false to skip over aged out data

crankycoder:
This enables skipping over data loss for the federated learning streaming jobs.

See bug https://bugzilla.mozilla.org/show_bug.cgi?id=1540562

### DIFF
--- a/src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
@@ -36,6 +36,7 @@ object FederatedLearningSearchOptimizer extends StreamingJobBase {
       .option("spark.streaming.kafka.consumer.cache.maxCapacity", 100)
       .option("subscribe", TelemetryKafkaTopic)
       .option("startingOffsets", opts.startingOffsets())
+      .option("failOnDataLoss", false)
       .load()
       .select("value")
 
--- a/src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
+++ b/src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
@@ -36,7 +36,7 @@ object FederatedLearningSearchOptimizer extends StreamingJobBase {
       .option("spark.streaming.kafka.consumer.cache.maxCapacity", 100)
       .option("subscribe", TelemetryKafkaTopic)
       .option("startingOffsets", opts.startingOffsets())
-      .option("failOnDataLoss", false)
+      .option("failOnDataLoss", opts.failOnDataLoss())
       .load()
       .select("value")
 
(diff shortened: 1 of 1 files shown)


src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
```diff
@@ -36,6 +36,7 @@ object FederatedLearningSearchOptimizer extends StreamingJobBase {
       .option("spark.streaming.kafka.consumer.cache.maxCapacity", 100)
       .option("subscribe", TelemetryKafkaTopic)
       .option("startingOffsets", opts.startingOffsets())
+      .option("failOnDataLoss", false)
```
akkomar: Can we instead use configuration option here to keep this consistent with other jobs?
You'd need to change this like:
https://github.com/mozilla/telemetry-streaming/blob/3c0a0015f91287b43743f0f2bae3b1b7d4addbde/src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala#L218
and add the option in this class' Opts` declaration, like:
https://github.com/mozilla/telemetry-streaming/blob/3c0a0015f91287b43743f0f2bae3b1b7d4addbde/src/main/scala/com/mozilla/telemetry/streaming/EventsToAmplitude.scala#L81-L82

Also: please [use the imperative mood in the commit message](https://chris.beams.io/posts/git-commit/#imperative)

crankycoder: Thanks @akkomar.  I've added d2e827f to add an option that defaults to false for this.

crankycoder:
Can you double check my work before I merge this?

@akkomar r?

src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
```diff
@@ -131,6 +132,9 @@ object FederatedLearningSearchOptimizer extends StreamingJobBase {
   }
 
   private class Opts(args: Array[String]) extends BaseOpts(args) {
+    val failOnDataLoss:ScallopOption[Boolean] = opt[Boolean](
```
akkomar:
```suggestion
    val failOnDataLoss: ScallopOption[Boolean] = opt[Boolean](
```

src/main/scala/com/mozilla/telemetry/streaming/FederatedLearningSearchOptimizer.scala
```diff
@@ -131,6 +132,9 @@ object FederatedLearningSearchOptimizer extends StreamingJobBase {
   }
 
   private class Opts(args: Array[String]) extends BaseOpts(args) {
+    val failOnDataLoss: ScallopOption[Boolean] = opt[Boolean](
+      descr = "Whether to fail the query when it’s possible that data is lost.",
+      default=Some(false))
```
akkomar:
```suggestion
      default = Some(false))
```

PR #236, opened 2019-04-17
Bug 1542349: Add new about:debugging devtools events to e2a configs

sunahsuh:
The Devtools team has instrumented new events for about:debugging and would like them in amplitude -- I've updated the configs for release and prerelease with them and tested in the devtools_dev project and all the events & properties shipped correctly as far as I can tell


sunahsuh:
Well, I restarted it twice and it took the second time, bleh.

I'll merge this in since the failure seems unrelated, and I'll file a bug about the flaky test behavior.

PR #242, opened 2019-05-21
Fire TV Events to Amplitude

fbertsch:
This follows the format of previous new pings. I kept device and arch optional, since they are included in the schema.

### DIFF
--- a/configs/fire_tv_events_schemas.json
+++ b/configs/fire_tv_events_schemas.json
@@ -1,0 +1,1545 @@
+{
+    "source": "telemetry",
+    "filters": {
+        "appName": [
+            "FirefoxForFireTV", 
+            "OTHER"
+        ], 
+        "docType": [
+            "mobile-event"
+        ], 
+        "os": [
+            "Android"
+        ]
+    },
+    "eventGroups": [
+        {
+            "eventGroupName": "Firefox for Fire TV", 
+            "events": [
+                {
+                    "amplitudeProperties": {}, 
+                    "description": "", 
+                    "name": "app - back", 
+                    "schema": {
+                        "$schema": "http://json-schema.org/schema#", 
+                        "properties": {
+                            "category": {
+                                "enum": [
+                                    "action"
+                                ], 
+                                "type": "string"
+                            }, 
+                            "method": {
+                                "enum": [
+                                    "page"
+                                ], 
+                                "type": "string"
+                            }, 
+                            "object": {
+                                "enum": [
+                                    "browser"
+                                ], 
+                                "type": "string"
+                            }, 
+                            "timestamp": {
+                                "minimum": 0, 
+                                "type": "number"
+                            }, 
+                            "value": {
+                                "enum": [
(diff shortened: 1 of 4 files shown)


fbertsch:
Tested on dev, here is an example chart: https://analytics.amplitude.com/mozilla-corp/chart/6ab6nnu

configs/fire_tv_events_schemas.json
```diff
@@ -0,0 +1,1545 @@
+{
+    "source": "telemetry",
+    "filters": {
+        "appName": [
+            "FirefoxForFireTV", 
+            "OTHER"
```
akkomar: Is "OTHER" needed?

configs/fire_tv_events_schemas.json
```diff
@@ -0,0 +1,1545 @@
...
+                        "properties": {
+                            "category": {
+                                "enum": [
+                                    "action"
+                                ], 
+                                "type": "string"
+                            }, 
+                            "method": {
+                                "enum": [
+                                    "click_or_voice"
```
akkomar: That's not fixable at this level, but I'm wondering if they'd want to know which one was that (click or voice).

configs/fire_tv_events_schemas.json
```diff
@@ -0,0 +1,1545 @@
...
+                                ], 
+                                "type": "string"
+                            }, 
+                            "timestamp": {
+                                "minimum": 0, 
+                                "type": "number"
+                            }, 
+                            "value": {
+                                "enum": [
+                                    "play"
```
akkomar: Typo? It's `"media - pause"` event.

```suggestion
                                    "pause"
```

configs/fire_tv_events_schemas.json
```diff
@@ -0,0 +1,1545 @@
...
+                                ], 
+                                "type": "string"
+                            }, 
+                            "timestamp": {
+                                "minimum": 0, 
+                                "type": "number"
+                            }, 
+                            "value": {
+                                "enum": [
+                                    "pause"
```
akkomar: Same as above?
```suggestion
                                    "play"
```

configs/fire_tv_events_schemas.json
fbertsch: Unfortunately it is, since we never moved `FirefoxForFireTV` out of the `OTHER` bucket

fbertsch: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1553585. Once we backfill and that is completed we can remove `OTHER`.

configs/fire_tv_events_schemas.json
fbertsch: Good question. Might be worth bringing to Ashley's attention.

configs/fire_tv_events_schemas.json
fbertsch: Good catch!

PR #248, opened 2019-07-01
Use driver hostname for datadog sink address in UptakeEventsToDatadog

sunahsuh:
Testing whether we can use the agent installed on the driver as the sink location


sunahsuh:
Nope, didn't work

PR #250, opened 2019-07-25
Bug 1566996: Add more devtools events

sunahsuh:
Added a bunch more devtools events. I'm hoping the next PR for devtools amplitude will be figuring out how to ship all of this over automatically via a bigquery query and csv export


sunahsuh:
No idea, but it seems likely we'll add more as we add more devtools features..

